PR #35, opened 2019-07-27
Fix typo, change from 6-bit to 8-bit(cliccfg)

lance91633:
It seems that cliccfg is a 8-bit memory-mapped register


Kevin-Andes:
Thanks for spotting this. This fix was incorporated in 74f86c38580b40e40738465f28c66dbfb6517774.

Issue #82, opened 2020-05-06
xcause registers


MoonbaseOtago:
I've been writing interrupt test code .... and it dawns on me that adding new non-0 bits to the xstatus registers may break existing trap handler code (for exceptions, not for interrupts)

It might make sense to create new xistatus registers that are the same as xstatus but with the new fields

(note: I seem to be making a lot of issues here, please poke me if I'm being too annoying)

Paul Campbell

brucehoult:
Hi Paul, I believe that would make such code already broken -- you're supposed to ignore reserved fields.

p.s. good to see you here. Let me know if you get to Wellington, once that is permitted...

MoonbaseOtago:
(yeah hi!) I guess it mostly depends on how existing code is written, at worst it's one extra instruction in the system call path

MoonbaseOtago:
BTW I guess that means the current linux entry.S is broken :-)

aswaterman:
Linux entry.S is actually a red herring, because it's writing sstatus from a previous value stored to memory.  The `start_thread` code that initializes that value in memory is the potentially problematic thing.  Point being, there's no need to add extra code to the critical code path in Linux.

It shouldn't be problematic in practice anyway because new bits will be added to sstatus in such a way as to not break existing software.

MoonbaseOtago:
sstaus is not the issue ... it's scause in:

handle_exception:
  .....
   csrr s4, CSR_CAUSE
   .....
   bge s4, zero, 1f
  .....
1:
  ....
  li t0, EXC_SYSCALL
  beq s4, t0, handle_syscall

The problem is that because we're adding scause.mpp/mpie/mpil then s4 needs to be masked before anyone checks it's value (the bge is OK though) - as i said it needs an extra masking operation before use

aswaterman:
The Linux SBI can (and should) mandate no-CLIC mode.

dansmathers:
FYI, this is related to issue #30 which asks when an exception is taken in CLIC mode, does mcause look like non-clic or clic.   This issue #82 assumes it looks like clic, but it should be clarified in the spec.

MoonbaseOtago:
Note: if you can't see the saved mcause.mpil you can't save it (is it still there or is it set to 0?) - this is of particular interest if you take a trap (for example a page fault) while accessing a clic vectored interrupt

kasanovic:
[...]

Issue #45, opened 2020-03-31
xtvec behavior when base and original mode both supported


JamesKenneyImperas:
Section 4.4 (New xtvec CSR mode for CLIC) implies that, when both original and CLIC mode are supported, it is possible to use original interrupt mode for one exception level and CLIC mode for another. For example, `mtvec` could be configured for CLIC mode and `stvec` for original mode. What happens if this is done?

As a contrived example, suppose that `mtvec` is configured to use CLIC mode, and the CLIC routes the Supervisor timer interrupt to Machine mode, but that `stvec` is configured to use original mode to route the same interrupt to Supervisor mode. Which interrupt controller wins?

MoonbaseOtago:
Here's a more detailed (and related) note I had added to a closed bug that maybe explains this another way:

One issue that probably needs to be clarified is what happens when a mode is running in CLIC mode and a lesser mode is not (for example if M-mode is using the CLIC but S-mode isn't) then the spec says that xdeleg of the mode in CLIC mode (mdeleg in the example above) is specified to not have any effect ... this means that interrupts can't be passed (to S-mode in this case).

I sort of see two possible solutions to this:

either say it's undefined and that if this situation arises it's not required to work, and suggest that if a mode uses the CLIC all lesser modes should too, or
say that xdeleg has no effect on mode-x, but does have the effect of delegating to the next lesser mode (for example mdeleg would not necessarily disable any M-mode interrupts, but it would cause them to be passed to S-mode)

kasanovic:
For first rev 1.0, we can require that all privilege modes can only run in same mode, i.e., only one of CLIC or CLINT is visible in lower privilege modes depending on M-mode setting.  This should be added to spec.

There is a covert channel concern if additional state (i.e., sstatus.sie) is visible to lower privilege mode but not context swapped by higher privilege modes, so higher privilege mode would have to be aware that it allowed lower privilege modes to switch interrupt mode.  There is a more general discussion about delegation of feature mode bits happening in hypervisor group.

kasanovic:
Closed by pull #128

kasanovic:
The text should state that only M-mode can select CLIC versus CLINT mode.  Lower privilege modes will not have the ability to change between CLIC and CLINT modes.

kasanovic:
[...]

PR #128, opened 2021-04-05
Update clic.adoc

dansmathers:
spec clarification for issue #45


dansmathers:
this spec clarification also should resolve #82 comment: For rev 1.0, we will mandate that all privilege modes see the same core-local interrupt controller (CLIC or CLINT), with only M-mode able to support configuring the interrupt controller.

Issue #125, opened 2021-03-22
Typo in definition of clicintattr[i].mode?


nati-rapaport:
clic spec, chapter 3.2.1 (Specifying Interrupt Privilege Mode) specifies: "Similarly, in systems that support all M/S/U-mode interrupts, cliccfg.nmbits can be set to 0, 1, or 2 bits to represent privilege-modes. cliccfg.nmbits = 0 indicates that all local interrupts are taken in M-mode. cliccfg.nmbits = 1 indicates that the MSB selects between M-mode (1) and U-mode (0)." etc.
On the other hand, the table below shows that when nmbits=1 and clicintattr[i].mode=0x, then the interpretation is S mode interrupt (see below). There is a contradiction here. So, who is right here - the wording or the table?

priv-modes  nmbits clicintattr[i].mode Interpretation
M                    0              xx                    M-mode interrupt
M/U                0              xx                    M-mode interrupt
M/U                1              0x                    U-mode interrupt
M/U                1              1x                    M-mode interrupt
M/S/U             0              xx                    M-mode interrupt
**M/S/U             1              0x                    S-mode interrupt**

dansmathers:
This looks like the same as issue #117.

Kevin-Andes:
Yes, this is an inconsistency and was also reported earlier in issue #117 as Dan mentioned. Sorry we did not have time to discuss this issue in our last week's meeting. We will try to discuss and resolve this in our next week's group meeting.

Thanks.

kasanovic:
Closed by pull #132

Issue #117, opened 2021-02-26
Section 3.2.1:  cliccfg.nmbits = 1 with M/S/U modes supported


konrad-schwarz:
The text says that in a system supporting M, S, and U modes, with nmbits set to one, msb clear results in the interrupt taken in U mode.  However, the table below shows the interrupt being taken in S-mode -- a much more logical and useful settting, IMHO.

Kevin-Andes:
Thank you for pointing out this inconsistency. According to the privilege spec (Table 1.2 in Sec. 1.2, supported combinations of privilege modes), it describes the supported modes are M and U if the number of privilege levels supported are 2. In this case, there is only 1 mode bits so the effective levels are 2.

We can discuss further which combinations are more common use cases: M plus U, or M plus S.

konrad-schwarz:
Unfortunately, I have scheduling conflicts during your meetings.

However, I’m not sure there is much to discuss here:

On a system supporting just M and U mode – which at least originally was the micro-controller mode, it is obvious that these two modes should be used.

On a system supporting M, S, and U mode – the standard “application processor” mode, required for running an operating system such as Linux, the ability to direct interrupts to S-mode – the mode the operating system is running on – is a non-negotiable requirement (otherwise interrupt driven IO is not possible).   The need for M-mode to handle interrupts, e.g., temperature warnings, fan settings, battery management, and so on, is well established industry practice; the analogue to M-mode on Intel x86 for example is System Management Mode (SMM) which uses a dedicated System Management Interrupt that has higher priority than any ordinary interrupt, and rightfully so, as ignoring or delaying handling of these interrupt requests can imperil the machine.

Kevin-Andes:
No problem, feel free to put down your comments/suggestions here and I can later bring them up in the meeting for discussion.

I see your points. For clarification, does Linux **only** need M-mode and S-mode interrupts typically? Or does Linux also need U-mode interrupts? If it needs all 3 modes then 2 bits must be used for privilege selection.

By the way, as an example, a potential use case of just M and U modes is when you use this processor as a micro-controller (even though it has capability of supporting M/S/U modes).

dansmathers:
This line is for a system has M, S, and U mode so if there is only 1 bit for nmbits, then it should select between M and S mode.

kasanovic:
[...]

PR #132, opened 2021-04-06
Update clic.adoc for issues #117, #125 change text to match table in M/S/U system if nmbits==1

dansmathers:
Spec update for issues #117, #125 Change text to match table in (Specifying Interrupt Privilege Mode).  in M/S/U system, if nmbits ==1 , selects between M-mode and S-mode.


Kevin-Andes:
Actually, this update may not be entirely correct. This is because CLIC is designed for small MCUs and they may use U-mode instead of S-mode more frequently. We still need to discuss and resolve this in our group meeting. 

Therefore, we may want to postpone this update and wait for our final resolution first.

dansmathers:
right.  I was just preparing the pull in case we decided it in the meeting.   I thought it might be the likely choice.
________________________________

PR #144, opened 2021-04-13
update WFI text

dansmathers:
fixes and simplifies changes attempted in rejected pull #135 and pull #121 .  closing pull #135 and pull #121


kasanovic:
Need to clarify that the machine-mode interrupt is not enabled, but the input can be asserted.  Level 0 is a locally disabled interrupt, but can still mask lower-mode interrupts, so don't use it to disable. Should use clicintie bits to locally disable interrupts and not use level 0.

PR #145, opened 2021-04-13
only m-mode can select CLIC vs CLINT mode

dansmathers:
spec update to close reopened #45

### DIFF
--- a/clic.adoc
+++ b/clic.adoc
@@ -732,13 +732,14 @@ power-of-two boundary.
 
 NOTE: Although future CLIC versions may allow privileges to have different 
 {tvec}.`mode` settings, for now all privilege modes must run in either CLIC mode 
-or all privilege modes must run in non-CLIC mode.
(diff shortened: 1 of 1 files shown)


Kevin-Andes:
Hi Dan,

Sorry, it is somewhat confusing what you meant by adding "xtvec (when mtvec.mode != 11)"...  If you are trying to describe old modes by excluding CLIC then this equation seems incorrect. For example, the behavior of mode "10" is also reserved and thus undefined yet. In addition, please note that we are only using a tiny fraction of mode "11" because we just used one value "6'b0000_11" while reserving other 15 values (i.e., 6'b0001_11 to 6'b1111_11) for future use. This is because we defined "In this mode, the trap vector base address held in
xtvec is constrained to be aligned on a 64-byte or larger power-of-two boundary."

Lastly, this line seems irrelevant to our issue subject "only m-mode can select CLIC vs CLINT mode"

Issue #141, opened 2021-04-13
Need to add commentary on use of bare-S mode (satp=0) versus proposed N-extension


kasanovic:
Bare S-mode (no MMU, satp=0) can be used in microcontrollers to allow hardware delegation of interrupts out of M-mode.   Bare S-mode has already been ratified as part of privileged architecture.  There are also proposals to add S-mode PMP support to allow an RTOS running in S-mode to isolate itself from tasks running in U-mode.  The proposed N-extension would also add user-mode interrupts and traps, but has not been ratified and is not currently being advanced.

This issue records that CLIC spec text should add context explaining this.

dansmathers:
created pull #146 to close this issue.

kasanovic:
Closed by #141

PR #146, opened 2021-04-13
bare-S mode comment for issue #141

dansmathers:


Kevin-Andes:
In our group meeting on 2021/04/13, we also mentioned that we should change the selection of U-mode from "0" to "reserved and requires N-extension".

dansmathers:
I think we decided to put the note describing that bare-S mode has been ratified but N-extension has not and is not being currently advanced.  I think in the meeting we decided not to remove u-mode from the spec but add the bare-S mode note because there would be too many changes to the spec to rip out u-mode interrupts.  We can discuss again

Kevin-Andes:
Agreed, I am not suggesting the removal of U-mode from the spec. I thought we agreed to just modify the encoding of U-mode to "reserved and requires N-extension" to avoid trouble in compatibility test.

Perhaps, I can later make this change in a separate commit.

PR #167, opened 2021-08-31
clicintctl definition simplification proposal

dansmathers:
proposal to simplify clicintctl to either represent interrupt levels (used to determine preemption for nesting interrupts) or priorities (does not affect preemption)


kasanovic:
Can improve text to describe as only 0 or 8 currently supported.

Issue #96, opened 2020-09-08
Proposed reformat of cliccfg


kasanovic:
From John Hauser, email March 10, 2019:
As currently defined, the CLIC has a single configuration byte,
'cliccfg', that determines the splitting of interrupt control settings
into levels and priorities.  This one byte applies to all of a hart's
privilege levels (M, S, and/or U) simultaneously.  I strongly believe
this is suboptimal.  Instead, each privilege level should be given
its own separate configuration.  As I'm sure you know, it's common for
each privilege level to be running software from different authors,
and while the different levels will often interact (through environment
calls and such), each privilege level should get to choose for itself
how many nested interrupt levels it must accomodate on its interrupt
stack, and whether to use selective hardware vectoring.  Such choices
should not be dictated from above, or from below.

Splitting up 'cliccfg' implies also an adjustment to the way the
'nmbits' field operates.  There will now be up to three configuration
registers, which here I will call 'clicmcfg', 'clicscfg', and
'clicucfg'.  The new registers can have these layouts:

  clicmcfg

  bits    field
  7:6     reserved (WARL 0)
  5       nmbits
  4:1     nlbits[3:0]
  0       nvbits

  clicscfg

  bits    field
  7:6     reserved (WARL 0)
  5       nmbits
  4:1     nlbits[3:0]
  0       nvbits

  clicucfg

  bits    field
  7:5     reserved (WARL 0)
  4:1     nlbits[3:0]
  0       nvbits

Bit 'clicmcfg.nmbits' determines whether any interrupts can be
delegated to privilege levels beyond M-mode.  If 'clicmcfg.nmbits' = 0,
then all interrupts are taken in M-mode, and 'clicscfg' and 'clicucfg'
are ignored.  Likewise, when S-mode is supported, bit 'clicscfg.nmbits'
determines whether interrupts can be delegated beyond S-mode to U-mode.
If 'clicmcfg.nmbits' = 1 and 'clicscfg.nmbits' = 0, then no interrupts
are delegated beyond S-mode, and 'clicucfg' is again ignored.

For illustration purposes, let's assume our CLIC implements all 8 bits
in each 'clicintctl[]' byte.  If 'clicmcfg.nmbits' = 0, then the 8 bits
are all used for the M-mode interrupt priority, split up by 'clicmcfg'
fields 'nlbits' and 'nvbits', same as before.  On the other hand,
if 'clicmcfg.nmbits' = 1, then bit 7 of each interrupt control byte
[...]

PR #172, opened 2021-09-27
pull request for issue #96 m/s/ucliccfg

dansmathers:
#96 – different priv modes may want different level/priority control.  mcliccfg, scliccfg, ucliccfg instead of single cliccfg.  Place at different byte offsets in case implementation wants to overlap m/s/u regions and control priv access in a method not using separate address ranges.
changed references to cliccfg to xcliccfg

### DIFF
--- a/clic.adoc
+++ b/clic.adoc
@@ -27,6 +27,8 @@
 
 :ret: pass:q[``**__x__**ret``]
 
+:cliccfg: pass:q[``**__x__**cliccfg``]
+
 :le: &#8804;
 :ge: &#8805;
 :lt: &#60;
@@ -325,7 +327,7 @@ M-mode CLIC memory map
   ###   0x00C0-0x07FF              reserved    ###
   ###   0x0800-0x0FFF              custom      ###
   
-  0x0000         1B          RW        cliccfg
+  0x0000         1B          RW        mcliccfg
   0x0004         4B          R         clicinfo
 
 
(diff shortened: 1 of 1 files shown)


kasanovic:
Discussed some changes to this in the TG meeting 2021/9/28:

Current design

----------------------------------------------------------------------
0x0000 xcliccfg
----------------------------------------------------------------------
0x1000 per-interrupt bytes
----------------------------------------------------------------------

One aperture per priv mode.

S-mode PMP NAPOT 0x0000-0x1FFF



Memory map for overlapping CLIC apertures, where assumption is bus
transactions carries privilege mode and CLIC interface modifies
behavior based on bus transaction privilege mode.

----------------------------------------------------------------------
0x0000 M-mode aperture
0x0000 mcliccfg
...   (reserved, future M-mode config registers)
----------------------------------------------------------------------
0x0400 HS-mode aperture
0x0400 hscliccfg
...
----------------------------------------------------------------------
0x0800 S-mode aperture
0x0800 scliccfg
...
----------------------------------------------------------------------
0x0C00 U-mode aperture
0x0C00 ucliccfg
...
----------------------------------------------------------------------
0x1000 per-interrupt bytes (all modes)
...
----------------------------------------------------------------------

S-mode PMP range  0x0800-0x1FFF (but intended for use with priv mode
on bus transactions)

Using shifted per-mode cfg with separate apertures:

----------------------------------------------------------------------
0x0000 mcliccfg
----------------------------------------------------------------------
0x1000 per-interrupt bytes (M-mode)

----------------------------------------------------------------------
...
----------------------------------------------------------------------
0x0400 HS-mode
----------------------------------------------------------------------
0x1000 per-interrupt bytes (HS-mode)

----------------------------------------------------------------------
...
----------------------------------------------------------------------
0x0000-0x07FF reserved
0x0800 scliccfg
----------------------------------------------------------------------
0x1000 per-interrupt bytes (S-mode)

----------------------------------------------------------------------
...
----------------------------------------------------------------------
0x0C00 ucliccfg
----------------------------------------------------------------------
0x1000 per-interrupt bytes (U-mode)

----------------------------------------------------------------------

kasanovic:
Discussion was that allowing mode-specific difference in functionality at same address would be problematic in general, and even in this specific case.  For example, an M-mode routine trying to access S-mode register would have to know to use mstatus.mprv feature to change effective mode on access.  For more general case in other specifications, we've adopted policy of not having mode-specific functional differences on allowed accesses to the same address.

dansmathers:
closing this.  created pull #183 instead.

Issue #48, opened 2020-03-31
Logic/state diagram for clicintip signals required


JamesKenneyImperas:
Section 3.4 (CLIC Interrupt Pending) attempts to describe the logic for `clicintip` signals, but I think this needs further clarification, either by a state diagram or a schematic. In particular, it says:

> `clicintip[i] `is a read-write register. Software-based (direct) writes to these pending bits have priority over hardware-based writes (triggers).

What does this mean if the input is configured as level-sensitive?

What state is latched and what is combinatorial? For example, what is the effect of reconfiguring a signal to be positive level-triggered instead of negative level-triggered while it is pending?

I think this section needs to fully describe the value of `clicintp` for every possible transition of the external input signal, corresponding `clicintie` and `clicintarttr` registers, and describe the effect of a software write to `clicintp` in every state.

anthony-coulter:
=== The CLIC triggering mechanism ===

The CLIC triggering mechanism for interrupt #i has the following components:

* The input signal from the interrupt source.
* Three configuration bits `trigger`, `polarity`, and `enable`
* An invisible storage bit `latch`
* The memory-accessible bit `pending`
* An input signal `autoclear` from the hart

The triggering mechanism thus latches four bits of state: `trigger`, `polarity`, `enable`, and `latch`.

=== Memory map of these bits ===

Four bits in the triggering mechanism are accessible via memory accesses:

* `trigger` is currently the second-least-significant bit of `clicintattr[i]`.
* `polarity` is currently the third-least-significant bit of `clicintattr[i]`.
* `enable` will probably be the least-significant bit of `clicintie[i]`.
* `pending` will probably be the least-significant bit of `clicintip[i]`.
* `latch` is not directly accessible in the memory map.

The value of the trigger bit is expressed with a symbolic constant: `LEVEL = 0` and `EDGE = 1`.

=== State of the bits on a reset ===

The state of the trigger, polarity, enable, and latch bits on a reset is unspecified. (The privileged spec already mandates that `mstatus.mie = 0` on reset, and that is enough to block all interrupts before they are configured by software.)

=== Reading these bits ===

Loads from the memory addresses corresponding to `trigger`, `polarity`, and `enable` read the saved bits as expected.

Loads from the memory address corresponding to `pending` returns the value:

        `pending := latch | (trigger == LEVEL & (source_signal ^ polarity)`

where `source_signal` is the signal from the interrupt handler.

[...]

PR #174, opened 2021-09-27
clarification for issue #48

dansmathers:
updated clicintip description to clarify only xnxti writes affect clicintip
added detail to mtvec section that edge-triggered interrupt pending bits are cleared by hardware after successful trap handler function address fetch.


dansmathers:
created a new pull #178 with same updates but with the merge issue fixed.

PR #178, opened 2021-11-08
clarification for issue #48

dansmathers:
updated clicintip description to clarify only xnxti writes affect clicintip
added detail to mtvec section that edge-triggered interrupt pending bits are cleared by hardware after successful trap handler function address fetch.
same updates as pull #174 but with merge issue fixed.

### DIFF
--- a/clic.adoc
+++ b/clic.adoc
@@ -689,9 +689,9 @@ by hardware interrupt inputs and by software.  The bit is set by
 hardware after an edge of the appropriate polarity is observed on the
 interrupt input, as determined by the `clicintattr[i]` field.
 Hardware clears the associated interrupt pending bit when an
-interrupt is serviced in vectored mode.  Software writes can set or
-clear edge-triggered pending bits, either directly by writes to the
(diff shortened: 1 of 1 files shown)


JamesKenneyImperas:
Hello Dan,

Can you give more detail about why the interrupt pending bit is cleared _after_ the successful trap handler fetch and not _before_? I don't think this feels right (or perhaps some extra detail is needed).

If the trap handler fetch fails, some other handler will run that will eventually return via the code snippet later in this section:

```
if (xcause.inhv) //xcause here refers to the cause CSR of the previous privilege mode
    pc := M[xepc]
else
    pc := xepc
```

This resumes execution at the trap handler address _but doesn't clear the interrupt pending bit_. This means the interrupt will trigger _again_ when the interrupt handler returns.

I think that either the interrupt pending bit should be cleared _before_ the trap handler fetch or there needs to be some extra logic to clear the bit in the first branch of the above snippet.

Thanks.

kasanovic:
The ip bit should be cleared once the hart has committed to taking the interrupt, which happens before table fetch.

PR #183, opened 2021-12-15
pull for #96 xcliccfg

dansmathers:
#96 adds separate mcliccfg/scliccfg/ucliccfg for priv specific control of clicintctrl settings.


dansmathers:
this pull is replaced by pull #231

Issue #193, opened 2022-01-19
Motivation of changes for xret instruction


divan0v:
Add, please, motivation of changes for 'xret' instruction behaviour.

Motivation for the hardware vectoring bit 'xinhv' is presented in section 5.4: 
> The hardware vectoring bit xinhv is provided to allow resumable traps on fetches to the trap vector table.

Could you explain and add similar note for 'xret' modification. From my opinion, exception on 
hardware vectoring operation is rare case and software could process such exception with old 'xret' 
behaviour manually to allow resumable trap without performance degradation. 

Thanks.

divan0v:
Thanks for commit creation. 

The following phrase describes privilege mode for operation M[TBASE + XLEN/8 * exccode]:
> For permissions-checking purposes, the memory access to retrieve the function pointer for vectoring is treated as
> a load with the privilege mode (also obeying MPRV and SUM bits) and interrupt level of the interrupt handler.

Could you also add phrase about privilege mode and interrupt level that is used for operation M[xepc]? Or phrase above covers both cases?

kasanovic:
Closed with #193 and #194.

Issue #154, opened 2021-04-28
xtval and xepc when access exception on the table load (section 5.4)


JamesKenneyImperas:
I don't understand the intended behavior for `xtval` and `xepc` as described in section 5.4. The specification says:

> If there is an access exception on the table load, both `xtval` and `xepc` holds the faulting address.

The subsequent clarification note then says:

> The interesting case is vertical traps, where a more privileged layer is handling page faults or other synchronous faults on vector table access. The regular code path in more privileged layer will want to use `xtval` to determine what bad virtual address to page in, but will not normally restore `xtval` when returning to faulting context (potentially after some time and other contexts have run) However, it will restore `xepc` (using `x` for more privileged mode here) before using `xret` on normal code path. This is a rationale for why both `xtval` and `xepc` are written with the faulting address.

If I read this correctly, this is saying that a return from a vertical trap will execute at an address in the trap vector table. This doesn't seem to make sense, because the trap vector table doesn't contain _instructions_; it holds _addresses_. There is another comment that:

> The `inhv` bit has no effect except when returning from a trap using an `xret` instruction.

However, I can't find any description of this behavioral effect - can you please say where this is described?

Thanks.

kasanovic:
The inhv bit when set at xRET informs hardware to repeat the table load using the address in xEPC to obtain the address of the trap handler that is then written to the PC instead of directly writing xEPC to the PC. i.e.,
```
if (inhv)
    PC = M[xEPC]
else
    PC = xEPC
```
One of the goals of this behavior is to avoid complicating the critical code paths for handling virtual memory in the more-privileged layer.  The more-privileged layer does not have to distinguish CLIC vector table reads from other forms of data page fault and can handle them using exactly the same code.

JamesKenneyImperas:
Hi Krste,

I understand the intent now.

Presumably the `inhv` bit is also cleared if and only if the `M[xEPC]` load is successful? The new pseudocode doesn't mention when this happens.

Thanks.

JamesKenneyImperas:
Also if the `M[xEPC]` load in an `xRET` is _not_ successful, what is the effect on processor state after the `xRET` instruction completes? An `xRET` would normally update various state bits (in `xstatus`, `xintstatus` and `tcontrol`, for example). Do any of these updates occur if the load fails?

JamesKenneyImperas:
[...]

PR #194, opened 2022-01-19
spec clarification for issue #193

dansmathers:
Adding note explaining motivation of changes for xret instruction using text from comments in issue #154.


divan0v (approved):

divan0v (approved):

divan0v:
Should 'xret' execution sets xcause.interrupt when xcause.inhv=1?
And unsuccessful vectoring clears xcause.interrupt to distinguish interrupt and exception?

dansmathers:
taking an exception during hardware vectoring is only recoverable if it occurs with different privilege levels.   
For example if there is a s-mode hardware vector and there is a page fault causing a m-mode exception.  So scause.interrupt is still set, scause.inhv is still set but mcause now indicates an exception.   Mret checks scause.inhv and considers mepc as an address instead of an instruction.  but scause.interrupt is still set so mret doesn't need to set it.   I agree that the use of the X in xcause and xret make this confusing.  We should discuss in the next TG meeting if there is a better way to express this.

divan0v:
Thanks for clarification. It's became clearer.

PR #204, opened 2022-03-01
pull for issue #96 privilege proposal

dansmathers:
created a pull for the proposal in #96 to use clicintctrl level values determine interrupt privilege mode instead of using separate clicintattr.mode bits determine the interrupt privilege mode.


dansmathers:
contains more than one fix.  closing.  will create a new one.

PR #206, opened 2022-03-04
add vector table note

dansmathers:
based on comments in https://github.com/riscv/riscv-code-size-reduction/issues/134:  Add note that FENCE.I after updates to vector table will be needed on implementations with separate caching structures.


dansmathers:
fixed with pull #227

PR #215, opened 2022-03-09
xscratchcsw pseudo-code clarification

dansmathers:
updating xscratch pseudo-code to be more SAIL-like and be more explicit in how xPP and cur_privilege affect xscratch.  It was not clear from the previous mscratchcsw pseudo-code how sscratchcsw behaves in the non-standard use-case of when sscratchcsw is accessed when cur_privilege mode is m-mode.

### DIFF
--- a/clic.adoc
+++ b/clic.adoc
@@ -2291,16 +2291,30 @@ instructions set both `rs1` = `sp` and `rd` = `sp`.
 
 [source]
 ----
-  csrrw rd, mscratchcsw, rs1
+csrrw rd, sscratchcsw, rs1
 
-  // Pseudocode operation.
-  if (mcause.mpp!=M-mode) then {
-      t = rs1; rd = mscratch; mscratch = t;
-  } else {
-      rd = rs1; // mscratch unchanged.
-  }
+match cur_privilege {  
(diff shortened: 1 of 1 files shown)


divan0v:
From privilege specification: 
_"The xPP fields can only hold privilege modes up to x, so MPP is two bits wide, SPP is one bit wide, and UPP is implicitly zero."_

It seems to me that sstatus.SPP could not be never matched with machine mode.

divan0v:
Ah, sorry for my previous comment. I understood SAIL syntax. Your pseudocode is good.

divan0v (approved):

kasanovic:
Machine mode should look at mstatus.mpp to find out the previous privilege level.  Although machine-mode access to sscratchcsw is not expected to be the normal usage, we should specify in a way that simplifies hardware.  Usually higher-privilege modes can access lower-privilege modes' CSRs.

kasanovic:
Need to clarify operation when MPP allows for hypervisor mode.
Also, need to clarify operation when MPP is two-bits wide, hypervisor mode is not implemented, but something has set MPP to hypervisor mode.

PR #217, opened 2022-03-09
allow implementing less than 8 bits for xintthresh

dansmathers:
Added WARL and restriction on which bits need to be implemented for xintthresh

### DIFF
--- a/clic.adoc
+++ b/clic.adoc
@@ -1294,7 +1294,8 @@ The interrupt-level threshold ({intthresh}) is a new read-write WARL CSR,
 which holds an 8-bit field (`th`) for the threshold level of the
 associated privilege mode.  The `th` field is held in the least-significant
 8 bits of the CSR, and zero should be written to the upper bits.
-If the number of bits actually implemented in this register is less than 8 (e.g. `CLICINTCTLBITS`), the implemented bits are kept left-justified in the most-significant bits of the 8-bit field, with the lower unimplemented bits treated as hardwired to 0 (to allow {intthresh} to be programmed less than `clicintctl[__i__]` values if `CLICINTCTLBITS` < 8). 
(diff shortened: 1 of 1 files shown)


tovine:
There is a problem with this suggestion, as it makes it impossible to block the highest level of interrupts when unimplemented bits are assumed 0.
In effect, interrupts with level 255 will always be above xintthresh as long as _any_ of the lower threshold bits are not implemented.

To make this work (i.e. making it possible to block all interrupts by writing all implemented bits to 1), `xintthresh` needs to implement _at least_ (`CLICINTCTLBITS` + 1) bits, and all unimplemented bits must follow the state of the the lowest implemented bit.
E.g. an implementation with `CLICINTCTLBITS=3` would look something like this:

| Implemented bits | Actual value | Effective level |
|-------------------|--------------|--------------|
| 1110xxxx             | 11100000    | 224               |
| 0000xxxx             | 00000000    | 0                   |
| 0001xxxx             | 00011111    | 31                 |
| 1111xxxx             | 11111111    | 255 (blocking all) |

Whereas with the current proposal it would be like this:


| Implemented bits | Actual value | Effective level |
|-------------------|--------------|--------------|
| 1110xxxx             | 11100000    | 224               |
| 0000xxxx             | 00000000    | 0                   |
| 0001xxxx             | 00010000    | 16                 |
| 1111xxxx             | 11110000    | 240 (*not* blocking all) |

Actually, to _really_ make it behave similarly to the interrupt level bits (where all unimplemented bits are assumed to be 1) it would probably have to be something like this:

| Implemented bits | Actual value | Effective level |
|-------------------|--------------|--------------|
| 1110xxxx             | 11101111    | 239               |
| 0000xxxx             | 00000000    | 0 (blocking none) |
| 0001xxxx             | 00011111    | 31                 |
| 1111xxxx             | 11111111    | 255 (blocking all) |

Then (as you can see from the table above), the reason for needing one extra `xintthresh` bit (compared to CLICINTCTLBITS) is to enable setting the threshold to 0, and to distinguish the second and third cases (where the second should allow all interrupts and the third should block ones with the lowest configurable level).

This comment ended up a bit longer and complicated than initially intended, but I hope it was still understandable. 😄

Issue #240, opened 2022-05-17
mscratchcsw defined using mcause.mpp instead of mstatus.mpp


pashenden:
The pseudocode definition of csrrw access to the mscratchcsw CSR uses mcause.mpp for the previous privilege level. However, if mtvec.mode is not CLIC mode, mcause.mpp does not mirror mstatus.mpp and is read as 0. Should the definition use mstatus.mpp instead of mcause.mpp, so that the operation is the same in non-CLIC mode?

kasanovic:
mscratchcsw should not exist in non-CLIC mode, need to clarify text

dansmathers:
closed by pull #258?  default text that CSRs added by CLIC mode appear zero unless specified elsewise.

kasanovic:
Issue addressed by above commits.

PR #241, opened 2022-05-18
for issue #240 - change pseudocode for mscratchcsw to use mstatus.mpp

dansmathers:
For issue #240 
change pseudocode to use mstatus.mpp instead of mcause.mpp so mscratchcsw works in non-CLIC mode as well.


dansmathers:
Closing.  CLIC CSRs not visible in non-CLIC modes.

Issue #221, opened 2022-03-29
Add read-only shortcut to pending interrupts?


dansmathers:
Currently to be able to see which interrupts are pending, software would have to do a LD instruction to access each clicintip.  
Should we add a read-only ip status range with 1 bit per pending interrupt (similar to AIA external interrupt-pending registers eip0-eip63) so software can quickly see the state of interrupt pending bits?

dansmathers:
It might also be useful to have a memory map view similar to AIA external interrupt-enable registers (eie0-eie63) where software can quickly enable, disable multiple interrupt enables.   xintthresh kind of allows this kind of ability, but sometimes the set of interrupt enables may be based more on functional mode instead of level priority.

dansmathers:
would it make sense to have a software clint compatibility mode option where writes to CSR xIP, xIE is equivalent to writing to these shortcut bits?  AIA directly references CLINT xIP xIDELEG registers.   might be nice to have CLIC compatible with software like riscv-sbi ( https://github.com/riscv-non-isa/riscv-sbi-doc/blob/master/riscv-sbi.adoc )which references xIP, xIE.  AIA compatible software will probably reference xIP, xIE also.  seems like it would be easy for CLIC to support this alternate access method for interrupt pending and interrupt enables.

kasanovic:
Not clear that CLIC needs this given that the hardware can automatically jump to the appropriate handler for a given interrupt input.  Adding ability to read some bits via CSR might be OK, but allowing writes requires that a CSR write generates a memory-mapped I/O transaction in most implementations.  Would need to see some use case that benefits from the feature.

dansmathers:
just pointing out that CLIC already has defined the ability to clear clicintip via CSR:
_If the pending interrupt is edge-triggered, hardware will automatically clear the corresponding pending bit when the CSR instruction that accesses xnxti includes a write_

kasanovic:
Having a compressed bitmap view of the ip and ie bits could be useful, but should be done via a memory-mapped access to the CLIC registers.  This would represent an alternative memory-mapped view of the same bits.
This should provide the needed functionality without requiring new CSRs that perform memory transactions.

PR #270, opened 2022-09-13
pull proposal for issue #221

dansmathers:
add mirrored view of clicintie/clicintip so software can quickly modify multiple interrupt enables for issue #221 

### DIFF
--- a/clic.adoc
+++ b/clic.adoc
@@ -388,7 +388,15 @@ M-mode CLIC memory map
   0x00B8         4B          RW        clicinttrig[30]
   0x00BC         4B          RW        clicinttrig[31]
 
-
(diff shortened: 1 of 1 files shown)


tovine:
I guess this also works with 64 bit accesses even though only 32 bit registers are specified (maybe there should be a note about this to remove ambiguity)?

I like the idea, not 100% sold on the name but I don't have a better alternative at the moment :)

dansmathers:
I was trying to match wording with AIA since it seems similar.
from aia spec:
"Registers eip0 through eip63 contain the pending bits for all implemented interrupt identities,
and are collectively called the eip array. Registers eie0 through eie63 contain the enable bits for
the same interrupt identities, and are collectively called the eie array."

Issue #248, opened 2022-06-21
Hypervisor mode should be discussed.


kasanovic:
Extension should either be incompatible with hypervisor extension, or behavior specified when hypervisor extension present.

kasanovic:
Hypervisor support is not included in current spec, will be added as a possible future extension.

kasanovic:
Reopening to tag as post v1.0 issue.

christian-herber-nxp:
Currently, the TGs view is that as fast interrupts is limited to single hart systems, we will not support H extension on the controller defined. Solutions with H extensions should use AIA as intended.
H and VS- specifications shall be removed from all extensions in the aclic document.

Issue #247, opened 2022-06-21
Behavior when MPP is two-bits wide and hypervisor not implemented


kasanovic:
Behavior is not specified in a few places (e.g., mscratchcsw).

kasanovic:
The plan is to leave hypervisor support out of current spec, to be added as a later extension at some later time.
The xPP registers are WARL, so cannot have an unsupported value, so it is OK that the pseudo-code does not describe the unsupported combinations.

PR #281, opened 2022-10-25
simple hypervisor proposal

dansmathers:
for issues #248 #247 .  Simple proposal.   using u-mode interrupt encodings with vs-mode interrupts when misa.H = 1.

### DIFF
--- a/clic.adoc
+++ b/clic.adoc
@@ -2705,6 +2705,158 @@ ID  Interrupt
 ----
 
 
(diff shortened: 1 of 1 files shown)


kasanovic:
Would not want to preclude the possibility of supporting user-mode traps, especially for system-management-like processes running alongside guest OSs.   Can change encoding to use the extra virtualized bit.

Issue #226, opened 2022-04-12
proposal to replace clicintattr.mode with xcliccfg.xlvl


dansmathers:
Separating out priv proposal from issue #96 
TG meeting discussed this proposal:

Replace nmbits and nlbits fields with:

mcliccfg.mlvl

scliccfg.slvl

xlvl values are encoded 0 ... 255 (only the most-significant
CLICINTCTLBITS are writable in the 8-bit field). Values are encoded
as for clicintctl (i.e., unimplemented LSBs are hardwired to 1s).

xlvl field values are absolute level numbers (not relative to mode).

mlvl is lowest level that traps into m-mode. If mlvl=0 all interrupts
trap into M-mode. If U-mode is not implemented (implying S-mode is
also not implemented), then mlvl=0.

If S-mode is implemented, slvl is lowest level that traps into s-mode.
mlvl has priority over slvl, so if mlvl<=slvl no interrupts trap into s-mode.
If slvl=0, no interrupts trap into u-mode. If U-mode interrupts are
not implemented, then slvl=0.

If U-mode interrupts are implemented, then interrupts <slvl trap into U-mode.

Operation:

The interrupt tree determines the highest-numbered pending-and-enabled
interrupts, based on value in clicintctl bits treated as an unsigned
integer.

If the value is >= mlvl, the trap is taken in M-mode.
otherwise if supervisor mode is implemented and the value is >= slvl, the trap is taken in S-mode
otherwise if user-mode interrupts are implemented the trap is taken in U-mode

Feb 28:
Is this an issue with using xlvl (I'm calling priv_thresh below) to set the privilege of the interrupt?

Since mcliccfg sets the level at which interrupts are m-mode, s-mode attempts to write clicintctl[__i__] above mcliccfg.priv_thresh should fail because we don't want s-mode to be able to change configuration of an interrupt to m-mode. But with attempts at writing clicintctl[__i__], SW can determine the higher privilege priv_thresh value. Is this a problem?

Mar 13
Benefits to proposal to replace nmbits and nlbits fields with mcliccfg.mlvl and scliccfg.slvl:

with m and 1 other mode supported, there is the possibility of saving1 implemented flop per interrupt (e.g. m has 4 and u has 12. no savings if m has 5 and u has 12. no savings if m has 8 and u has 8.
with m and 2 other modes supported, save at least 1 implemented flop per interrupt and it is possible to save 2 flops per interrupt if the packing works out right.
[...]

Issue #171, opened 2021-09-16
Inconsistency in Parameters


billhuffman:
Chapter 6 has a parameter, CLICCFGLBITS that says it's 0-CEIL(lg2(CLICLEVELS)) and that it's the number of bits implemented for cliccfg.nlbits.  This ties cliccfg.nlbits and the number of levels available.   But at the Sept 14 meeting, we said that nlbits could be hardwired to 0 or 8 and that 8 allowed any number of levels, not just 256.  If the Sept 14 discussion is correct, this parameter needs updating.

dansmathers:
closed bvy #369

Issue #158, opened 2021-06-09
CLICINTCTLBITS May Not Always Be Fixed


billhuffman:
Section 3.7 says that CLICINTCTLBITS is a fixed parameter.  But I think it may vary.  If cliccfg.nmbits is writable, then when it changes, the number of bits in the interrupt priority tree allocated to the mode in which the interrupt should be taken will vary.  CLICINTCTLBITS will then vary when cliccfg.nmbits is changed to utilize the remaining bits for level and priority.

dansmathers:
CLICINTCTLBITS is supposed to represent the number of implemented bits in clicintctl[i].  E.g, some smaller implementations may only want to have 4 flops per interrupt allocated to selecting level/priority instead of the full 8.  
The spec just describes when there are fewer than 8 bits implemented what the possible level/priority values the hardwired values take.   
there can't really be a dynamic logic area saving since this is fixed at implementation time.    

Are you thinking of an adjustable ctlbits option for a power saving option where the clic user could dynamically adjust which bits the priority encoder compares?   So the area implemented is the max 8 flops per interrupt but different use cases could attempt to reduce the dynamic power of the priority tree decoding?

billhuffman:
Yes, it is the number of bits implemented in clickintctl[i].  But that can vary.  For simplicity, let's assume no priority.  Just privileges and levels.  And let's assume we have four flops per interrupt and four-wide comparators in the tree.

If cliccfg.nmbits=0 as it is at reset, we have 15 interrupt levels available.   CLICINTCTLBITS = 4.

If we then write cliccfg.nmbits=1, one of the flops and comparator bits is used for the privilege the interrupt will be taken into.  That leaves three bits and allows 7 levels of interrupts.  CLICINTCTLBITS=3.

If we require CLICINTCTLBITS=3 in both cases, we restrict the use of the flops and comparator bits in an unfortunate way.

dansmathers:
I implemented it as a comparison of {priv_of_interrupt,cliccintctl[CLICINTCTLBITS implemented]}
so cliccfg.nmbits only affects the value of priv_of_interrupt.   
e.g, nmbits=0 then priv_of_interrupt is 2'b11 (machine).  nmbits=1, priv_of_interrupt is either 2'b00 (user) or 2'b11 (machine) based on cliccintattr.mode setting.
So the priority tree comparison is always 2+CLICINTCTLBITS wide

Here is the CLIC doc section I referenced for this:
[...]

Issue #49, opened 2020-03-31
Clarification of memory map required


JamesKenneyImperas:
I think section 3 (CLIC Memory-mapped Registers) requires clarification. It would be helpful to add diagrams of the entire address map for M,U and M,S,U cases so that it is clear how these regions abut. A diagram is also needed for a multi-hart system. This information is implied in words, but a diagram would be better.

In a multi-hart system, how does a hart programmatically determine the address of its S and U regions? This requires knowledge of the number of harts in the system.

JamesKenneyImperas:
Additionally, could you please clarify whether the behavior of memory-mapped accesses is dependent on the hart execution mode or not? For example, does an access by a hart executing in Machine mode to a Supervisor mode memory-mapped register behave exactly in the same way as a Supervisor mode access? As a second example, suppose that PMP settings are misconfigured to allow User mode access to the Machine mode control block. Would such an access behave in the same way as a Machine mode access to that block?

I ask this because in some other similar systems (e.g. ARM memory-mapped interrupt controllers) both the privilege mode of the block _and_ executing processor are taken into account. It looks like this is not the case here, but I wanted to check.

kasanovic:
This is generally a platform issue.   The spec itself focuses on the structure within each aperture not the location of the apertures in the global physical memory space.  Each platform needs to define a discovery mechanism to determine the memory map locations.

We should remove or downgrade to commentary the recommendation on the layout of multiple harts and multiple privilege modes, as different platforms will want to handle these differently.

kasanovic:
We added text back in #157 to explain the different privilege-mode memory mode apertures.  Please reraise issue if this is still not clear.

JamesKenneyImperas:
I've only just got round to looking at this (sorry) and do have another question about the memory map.

In older versions of the specification, pages for a multiple-hart system were laid out so that there was an initial control page (containing `cliccfg` etc), then four M-mode pages for each hart, then four S-mode pages for each hart, and so on:

```
    control page 
    M pages x 4 (hart0)
    M pages x 4 (hart1)
    ...
    M pages x 4 (hart N)
    S pages x 4 (hart0)
    S pages x 4 (hart1)
    ...
    S pages x 4 (hart N)
```

This explicit layout has now been removed. Can I confirm that whereas previously there was a single control page for all harts in a system, now there will be a control page _per hart_, with each hart having a different `mclicbase`, like this:

[...]

PR #286, opened 2022-11-03
pull to allow for multiple types of clicintattr.mode implementations

dansmathers:
pull for issue #96 #158 #171 #226 #49 
instead of defining number of bits implemented in clicintattr.mode, define parameters so that software knows valid values to program for clicintctl and any clicintattr.mode side effects.  this allows for additional methods of clicintattr.mode implementations while maintaining compatibility with previously defined method


dansmathers:
closed due to pull #369

Issue #320, opened 2023-04-05
clarify interaction of CLIC with RNMI


JamesKenneyImperas:
Section 4.10 (Resumable Non-Maskable Interrupts) says:

> The resumable NMI (RNMI) extension adds additional mnepc, mncause, and mnstatus CSRs. When in CLIC mode, *the RNMI retains the CLIC’s additional state in the mncause and mnstatus CSRs* to allow CLIC interrupts to be resumed.

Please can you make it explicit exactly what state is retained here and restored by MNRET?

Thanks.

kasanovic:
It is the same addition as to the original CLINT cause and status, but we can add explicit text for this.

dansmathers:
fixed will pull #328

PR #328, opened 2023-04-20
RMNI text clarification for issue #320

dansmathers:
added explicit text detailing CLIC fields added to RNMI CSR mncause for issue #320


dansmathers:
do we need to add text about RNMI NMRET and clic interrupt level?

Issue #314, opened 2023-03-26
smclicshv - Software vectoring only allowed when hardware vectoring is disabled?


jgonion:
Shouldn't the HW & SW vectoring schemes be consistent?

The avoidance of context save/restore at the start/end of interrupt handlers is valuable regardless of whether hardware vectoring is enabled or not.

Currently, mnxti only allows software vectoring wwhen hardware vectoring is disabled (clicintattr.shv == 0)

dansmathers:
7.2.5. smclicshv Changes to Next Interrupt Handler Address and Interrupt-Enable CSRs (xnxti)
adds the following requirement to xnxti csr:   we will discuss this in TG.

&& (clicintattr.shv==0) ) {
   // There is an available, non-hardware-vectored interrupt.

kasanovic:
This is to allow hardware and software handlers to have different software interfaces.  The assumption is that hardware vectoring would have customized save/restore, whereas the software vectoring uses a generic save/restore.

efoster029:
The real problem is the smclicshv extension currently only allows software or hardware vectoring, not both, which is a shame. 

If hardware vectoring used XTVEC and software vectoring used MTVT, both hardware and software vectoring could live harmoniously, giving RISC-V the best of both worlds: Hardware vectoring for higher priorities avoiding the overhead of software dispatch, and software vectoring horizontal priorities, avoiding the overhead of unnecessary context save/resore.

The current definition allows us to select between two unwanted taxes: People either need to pay the extra overhead of software getting involved in every interrupt dispatch (SHV=0), or they have to pay the tax of unnecessary context save/restore (SHV=1).

If we use XTVEC for hardware vectoring (as RISC-V intends) and XTVT for horizontal software vectoring, everything just works, and we would have a fundamentally more capable platform, that has distinct interfaces for HW vs SW vectoring while also allowing both to co-exist.

kasanovic:
In current  design, individual interrupts can be selected to be vectored in hardware or software (setting of clicintattr.shv), so we're unclear what is meant by "currently only allows software or hardware vectoring, not both" in issue above.

The same xtvt table is used for both software and hardware vectoring, but the software handler is located at xtvec.

In current design, xtvec points at instructions (as with CLINT style interrupts and regular RISC-V exceptions), whereas xtvt points at table entries containing function pointers.   xtvec is where the software handler is located, and xtvec is also used for exception handlers so cannot be repurposed for a software vectoring table.

[...]

PR #342, opened 2023-07-14
add mtvec submode for switching between vectored handlers

dansmathers:
for issue #314, add mtvec submode to allow dynamic switching between vectored handlers.  e.g. vectored handlers that save/restore context and vectored handlers that do not save/restore context.


dansmathers:
closing.  will try to create a method that doesn't double the size of the vector table.

Issue #308, opened 2023-03-03
Any higher priority interrupt causes xnxti service loop exit


MarkHillHuawei:
We have been studying the reference 11.1 C-ABI Trampoline Code and noticed that any pre-emption that occurs during the service_loop for horizontal interrupts will inevitably leave mcause.pil raised to the current interrupt level when it resumes. Then, when the mnxti at step (12) executes, the check that clic.level>mcause.pil will fail even when there are interrupts to service at the current level (or any lower level which is above the value of mcause.pil on entry to the CSR). The service loop will therefore exit and an unnecessary restore and save sequence initiated.

We see a few options:
1) Decide that was "unlucky" and hope the situation does not occur frequently enough to care, not sure that holds in an embedded highly interrupt driven system.

2) Immediately after the C ABI Routine returns at step (11), and prior to the mnxti at step (12), reload the value of mcause saved on entry,  disable interrupts and then restore the value of mcause to the CSR, then proceed to step 12 (which will re-enable interrupts). Disabling the interrupts ensures that no pre-emption can occur between the restoration of mcause and the mnxti. However, this costs two CSRs and a load so will add a few cycles to every iteration of the service loop.

3) Define a behavior for xnxti when the the source operand is a register rather than an immediate. This would always set the enable bit but would check the clic.level against bits 23:16 of the source operand instead of relying on xcause.xpil. The service_loop then passes in the value mcause had on ISR entry. This approach avoids the need for the two CSRs used in Option 2 to disable interrupts and restore mcause.

It is also possible to remove the load from the service loop by keeping the mcause in a callee-save register but this does mean an additional register needs to be saved and restored on entry/exit to the ISR so may not be a "win"

BTW I have also noticed a lack of consistency about whether it is mcause.mpil/xcause.xpil (10 times) or mcause.pil/xcause.pil (13 times). Probably worth tidying up and settling on one or the other. I have no preference.

MarkHillHuawei:
Option 4!

Define a behavior for xnxti when the the source operand is a register rather than an immediate. This would always set the enable bit but would check the clic.level against bits 23:16 of the source operand instead of relying on xcause.xpil. The service_loop then passes in the value mcause had on ISR entry.  This approach avoids the need for the two CSRs used in Option 2 to disable interrupts and restore mcause. 

[...]

PR #343, opened 2023-07-14
add CSRRW option for xnxti CSR

dansmathers:
for issue #308, allows setting mcause.mpil (used by xnxti to filter interrupts) as atomic with xnxti write.


kasanovic:
The new form should not copy register source [4:0] into low bits of mstatus - we need to set the interrupt enable bit.
Could use a csrrs instruction that checks level field in source register and always only sets the interrupt enable bit.  This is slightly unusual csr instruction, but does not take any more encoding space as it was one of the previously unused forms of accessing the mnxti CSR.

PR #344, opened 2023-08-15
add csrrs mnxti behavior

dansmathers:
For issue #308, allows using CSRRS mnxti to compare valule in rs1 against clic.level instead of mcause.pil which could have been modified by pre-emption.  (replaces pull #343)


MarkHillCodasip:
Thanks for addressing this issue (which I'm returning to as my new colleagues were confused about the purpose of the xnxt csrrs variant). It would be useful to have some supporting text to explain why the instruction picks out bits [23:16] from rs1 and to update the example C-ABI trampoline code to use this variant of the CSR. Let me know if it would be better to raise a new issue, I thought not as it is a clarification not a normative change.

Issue #347, opened 2023-08-18
exception_handler pseudocode needs checking


JamesKenneyImperas:
It looks to me that the `exception_handler` pseudocode in section 7.2.1 does not include the `XLEN/8` forced alignment that was introduced in April 2023. Could you please check and correct if necessary?

I think there is also a missing brace, inconsistent syntax and incorrect indentation on the line:
`    if (check_fetch_permissions(xepc) = Addr_OK)
`
Should this be:
`      if (check_fetch_permissions(xepc) = Addr_OK)
      then {
`
or something like that?

Thanks.

PR #354, opened 2023-08-29
fix exception_handler pseudocode

dansmathers:
For issue #347, add XLEN/8 forced alignment and fix missing brace in pseudo-code


kasanovic:
For clarity Sail pseudo-code should calculate masked epc once in inhv then clause, then use this value for both validity check and instruction fetch.

dansmathers:
closing.  attempting a better pseudo-code fix with pull #363

Issue #349, opened 2023-08-18
Moving CLIC memory mapped registers to indirect CSRs


christian-herber-nxp:
I wanted to bring up the possibility of migrating the memory mapped registers to indirectly accessed CSRs.
I assume that the reason for using memory mapped registers was the number of registers being too high for regular CSRs, but I am happy to learn from somebody with a longer history on that.

With the indirect CSRs on the way, I would think this is a nicer way of integrating those registers, which would also be more consistent with AIA. Also, with well defined CSR indeces, you no longer have to discover the base addresses of the CLIC memory regions.

Is this something which could be considered?

dansmathers:
adding link to indirect CSR github: https://github.com/riscv/riscv-indirect-csr-access

dansmathers:
Discussed in last 10 minutes of 8/29/2023 meeting. Probably something that should be discussed more.

dansmathers:
Pros: 
Similar to AIA.  ton of registers, nice to have as CSRs but don't want to use limited CSR space.  indirect CSRs opens up this option.
don't like to use pmp to separate clic memory regions since pmp regions are expensive.
easier to implement indirect CSRs in SAIL/spike?  
can limit corner cases vs. using memory interconnect?
for software consistency/debugger could remove CLIC base addr from parameter and have it be the same between implementations.  stability/portability.

Cons:
probably delay path to ratification?  already some implementations using memory mapped solutions?  could we have a way to coexist?  

sub extension to clic to use indirect CSRs?  single way is best for one piece of reference software.

if we were to implement, would we pack the enables or keep the same memory format?  CSR instructions allow setting easily.  configuring at system startup if packed would be easier.

christian-herber-nxp:
One question I have around the indirect CSRs is related to the user registers.
The indirect CSR access is only defined for M and S mode. CLIC is sort of an outlier with User mode support. Could this be an issue?

christian-herber-nxp:
This is a proposal for a potential CSR map for CLIC using S*csrind.

- I would expect the CSRs which are currently directly accesible to stay unchanged.
- xcliccfg is a single 32-bit register, can be mapped to a single indirect CSR
- clicintip and clicintie should be collapsed into individual bits of CSRs, analogous to AIA
- clicinttrig[0-31] will be indirectly accessed CSRs
[...]

Issue #225, opened 2022-04-11
Only require that effects of stores to clicintie/clicintip are evaluated within bounded time (not requiring immediate effects)


Silabs-ArjanB:
RISC-V Privileged specification 1.12 states the following about the MIP and MIE csrs:

> These conditions for an interrupt trap to occur must be evaluated in a bounded amount of time
> from when an interrupt becomes, or ceases to be, pending in mip, and must also be evaluated
> immediately following the execution of an xRET instruction or an explicit write to a CSR on which
> these interrupt trap conditions expressly depend (including mip, mie, mstatus, and mideleg).

Can a similar statement(s) be added in the CLIC spec related to clicintie and clicintip? I am particularly looking for a statement saying that determination of whether an interrupt is 'pending and enabled' must be evaluated in a bounded amount of time and that there is **no** requirement that an explicit (store) write to clicintie or clicintip are evaluated immediately.

Note that for CSR write (as for MIP/MIE) it is relatively easy to guarantee interrupt traps on the instruction immediately following such a CSR write. If similar requirements would exist on stores to clicintie/clicintip, then this would seriously complicate designs. The core normally would not even know that a write is to clicintie/clicintip. If a requirement for 'immediate evaluation' would exist, then it would also seriously lower performance as loads and stores can likely no longer be performed in a back-to-back manner when using protocols as AHB. It would also complicate for example pipelining the arbitration tree logic.

kasanovic:
Agree.

kasanovic:
Change  write-> store to be clearer bound is relative to an instruction execution in same program stream.

PR #364, opened 2023-10-10
change access to clic registers from memory mapped to indirect CSR

### LINKED ISSUE #221: Add read-only shortcut to pending interrupts?

dansmathers:
for issue #349, add option for using indirect CSR access method to control clic registers.  only updated text for smclic.  will update for text other priv modes later. Adding indirect CSR access could close #221 issue as well as it would provide a method to quickly enable/disable sets of interrupt enables.


clic.adoc
```diff
@@ -387,9 +387,45 @@ interrupt input.
...
+  ...
+  0x880     clicintie[XLEN-1:0]       
+  0x881     clicintie[2*XLEN-1:XLEN]  
+  ...
+  0x900     clicinttrig[0]
+  0x901     clicinttrig[1]
+  ...
+  0x91F     clicinttrig[31}
+  0x920     reserved for smclicconfig extension
+----
```
christian-herber-nxp, line 420: It seems that registers are only ever mapped to mireg, and mireg2-6 are unused. For configuration, that would cause quite some overhead. My suggestion would be to extend the table with five additional columns for mireg2-6, and to map registers there.

clic.adoc
```diff
@@ -387,9 +387,45 @@ interrupt input.
...
+  0x800     clicintip[XLEN-1:0]       
+  0x801     clicintip[2*XLEN-1:XLEN]  
+  ...
+  0x880     clicintie[XLEN-1:0]       
+  0x881     clicintie[2*XLEN-1:XLEN]  
+  ...
+  0x900     clicinttrig[0]
+  0x901     clicinttrig[1]
+  ...
+  0x91F     clicinttrig[31}
```
christian-herber-nxp, line 418: ends in a curly brace

clic.adoc
```diff
@@ -387,9 +387,45 @@ interrupt input.
...
 == smclic M-mode CLIC extension
 
-=== CLIC Memory-Mapped Registers
+Implementations can provide access to CLIC registers clicinttrig[i], clicintip[i], clicintie[i], clicintattr[i], clicintctl[i] 
+by memory-mapped registers or by using the Indirect CSR Access method (Smcsrind/Sscsrind).  
+The clic register accesses method available to software is indicated by the parameters CLICMEMMAPPED and CLICCSRINDIRECT.
+
+=== CLIC Register Access via Indirect CSR Access 
+
+The Indirect CSR Access allocated range for CLIC is TBD.  These address provided below are relative to the base of that allocated range.
```
christian-herber-nxp, line 396: Spurious space after TBD.
I think it is clearer to refer to select values rather than addresses. Select values are a kind of address, but its a more precise term in this context.

christian-herber-nxp (approved):
looks good to me

clic.adoc
kasanovic: We need to keep the indexing straightforward for software.  The only resource being consumed is miselect address space, and there is not that much being used, since there are 32 interrupts at each index.

clic.adoc
```diff
@@ -377,94 +377,144 @@ The existing timer (`mtip`/`stip`), software
...
+[%autofit]
 
-NOTE: A bus memory map or locked PMP entries could prevent M-mode
-software on a particular hart from reaching the CLIC memory map.
+----
+ miselect  |     mireg     |  mireg                      |     description    |
+ offset    |     bits      |  register state             |                    |
 
-NOTE: For reserved memory regions, specific trap behavior is not specified. Depending on system bus architecture, the system can ignore the access (e.g., read zero/write ignored) or cause a bus error (usually imprecise interrupt), or some other platform-specific behavior. The "reserved" annotation here implies that future standards might place additional standard registers in that space, and so using the space for non-standard features is inadvisable.
+  0x0      |     31:0     |   R or RW   clicintip[31:0]  |  clic interrupt pending for interrupts 31-0 |
```
kasanovic, line 402: Should just mark as RW, as we can ignore writes if the interrupt is not edge-triggered.  This way we don't have to try to deal with exceptions on writes to read-only bits in a register.

clic.adoc
```diff
@@ -525,11 +575,9 @@ or 255.
...
 
 Each interrupt input has a dedicated interrupt pending bit
-(`clicintip[__i__]`) and occupies one byte in the memory map for ease
-of access.  The pending bit is located in bit 0 of the byte.  Software
-should assume `clicintip[__i__]=0` means no interrupt pending, and
-`clicintip[__i__]!=0` indicates an interrupt is pending to accommodate
-possible future expansion of the `clicintip` field.
+(`clicintip[__i__]`). 
+Software should assume `clicintip[__i__]=0` means no interrupt pending, and
+`clicintip[__i__]!=0` indicates an interrupt is pending.
```
kasanovic, line 580: Now IP is only a single bit, so can reword description.

clic.adoc
```diff
@@ -562,8 +610,8 @@ level-sensitive mode.
 
 === CLIC Interrupt Enable (`clicintie`)
 Each interrupt input has a dedicated interrupt-enable bit (`clicintie[__i__]`)
-and occupies one byte in the memory map for ease of access. This control bit is
-read-write to enable/disable the corresponding interrupt.  The enable bit is located in bit 0 of the byte. Software should assume clicintie[i]=0 means no interrupt enabled, and clicintie[i]!=0 indicates an interrupt is enabled to accommodate possible future expansion of the clicintie field.
+This control bit is read-write to enable/disable the corresponding interrupt.  
+Software should assume clicintie[i]=0 means no interrupt enabled, and clicintie[i]!=0 indicates an interrupt is enabled.
```
kasanovic, line 614: Same comment on IP now being one bit.

Silabs-ArjanB:
@dansmathers The PR title and some of the text in this PR seems to suggest that an _option_ is being added. Reading through the PR it however seems that this is not the case because:

- The original _memory mapped option_ seems completely removed.
- No parameter for this option seems to have been introduced.

There is a statement that _"A CSR accessible via one method may or may not be accessible via the other method such as memory-mapped access"_. This _"other method"_ seems however not specified. 

I would really prefer if the memory mapped options stays explicitly specified.

The indirect CSR access option also opens up the discussion again that was closed here: https://github.com/riscv/riscv-fast-interrupt/issues/225. For indirect CSR accesses will there now again be requirements that writes to clicintie or clicintip are evaluated immediately? I think the answer to this question should be made explicit. If the answer is 'yes' then that introduces a weird incompatibility between the two access methods.

dansmathers:
Good question on should both indirect and memory mapped access be in this pull.  I've updated the pull title to reflect the current pull status. originally the pull just added indirect CSR access but you are correct that the current pull now replaces memory mapped access with indirect CSR access.   Hopefully we can have a good discussion on adding the memory mapped method back into the spec and how it should be, e.g. if it would be an extension or parameter or something else.

in the 20231205  fast interrupt TG minutes there was this discussion: 
> Issue #349 – using indirect CSRs for clicintctrl/attr/ip/ie. Hesitate to implement both indirect and CSRs and memory mapped.  choose one or the other. 
AIA – one register to implement.  try to be compatible with AIA?  in terms of msi – keep that consistent.  restrict memory map for software interrupts.  just because have msis, don’t open up the whole memory map to everything. 
Don’t' have to have indirect csrs.  understanding indirect csr as base.  memory map just opened up enough to do software interrupts. 


for your question about issue #225:
I still see this statement under clicintip section:
> The conditions for an interrupt trap to occur must be evaluated in a bounded amount of time from when an interrupt becomes, or ceases to be, pending in clicintip, but unlike the MIP/MIE CSRs, there is no requirement that clicintie or clicintip are evaluated immediately following an explicit store to clicintip or clicintie.

in the 20230926 fast interrupt TG minutes, there was this discussion:
> #349 - Fixed csr so pipeline can figure out conflicts.  indirect csrs, don’t know select register.  value may not be ready yet.  so things in ind csrs can’t cause pipeline events.  so means ind csrs aren’t critical timing.  so flexible but more constrained on what they can do.  just like AIA, it is a good candidate.  so ind csrs is like a local memory map

So for this question:
>  For indirect CSR accesses will there now again be requirements that writes to clicintie or clicintip are evaluated immediately?

I would say the answer is no and the answer is explicit in the spec/pull already.  Do you agree?

Silabs-ArjanB:
> for your question about issue https://github.com/riscv/riscv-fast-interrupt/issues/225:
> I still see this statement under clicintip section:
> 
> The conditions for an interrupt trap to occur must be evaluated in a bounded amount of time from when an interrupt becomes, or ceases to be, pending in clicintip, but unlike the MIP/MIE CSRs, there is no requirement that clicintie or clicintip are evaluated immediately following an explicit store to clicintip or clicintie.

This section was added because clicintip and clicintie are/were memory mapped. If you change these to be CSR accessible then I think this should be removed and the reverse should be mandated (i.e. they must be evaluated **immediately** following the execution of an xRET instruction or an explicit write to these CSRs).

I think furthermore that mandating indirect CSR access just increases the interrupt latency without any real benefit over using a memory mapped strategy. The reason I think interrupt latency is negatively affected is that the mireg CSR adds state that now needs to be saved and restored (which conflicts with the goal of having an extension that improves interrupt latency).

My proposal would be to mark indirect CSR access as a possible post-v1.0 addition as opposed to a pre-v1.0 replacement of a memory mapped strategy.

christian-herber-nxp:
Interesting point regarding the additional state. In most cases, however, I would argue that ISRs would not need to access indirect CSRs, and then no state needs to be saved.

Given the fact that indirect CSR accesses are a possibility now, I think they are the best choice for CLIC, and we want to have the best possible version of CLIC.

Of course I understand that it might be desired to maintain the memory mapped option in the interest of existing implementations, but from software point of view that would essentially lead to two kinds of CLICs (fueling any concerns regarding fragmentation).

I am travelling on Monday and will not be able to join, so please consider this as my input to the discussion.

dansmathers:
I tried to copy comments about moving CLIC memory mapped registers to indirect CSRs to issue #349 for easier issue tracking.  Can we continue the discussion there?

PR #370, opened 2024-02-10
change xnxti to also return SHV trap-handler entries

dansmathers:
for issue #314 
added note that software may choose to avoid unnecessary context save/restore in hardware-vector handlers by checking mcause.mpil

### DIFF
--- a/clic.adoc
+++ b/clic.adoc
@@ -1878,55 +1878,15 @@ Note: For now all privilege modes must run in either CLIC mode or all privilege 
 In CLIC mode, when a trap is taken, {cause} has the CLIC format and the {cause} fields are updated.
 On the other hand, when not in CLIC mode, {cause} has the CLINT mode format.
 
-==== smclicshv Changes to Next Interrupt Handler Address and Interrupt-Enable CSRs ({nxti})
+==== smclicshv and the Next Interrupt Handler Address and Interrupt-Enable CSRs ({nxti})
 
-A read of the {nxti} CSR using CSRR returns either zero, indicating there is no
-suitable interrupt to service or that the highest ranked interrupt is
-SHV or that the system is not in a CLIC mode, or returns a non-zero
-address of the entry in the trap handler table for software trap
-vectoring.
+With the smclicshv extension, if the highest ranked interrupt is SHV, {nxti} returns a pointer to the hardware-vectored trap-handler entry. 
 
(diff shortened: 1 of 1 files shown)


clic.adoc
```diff
@@ -1878,55 +1878,15 @@ Note: For now all privilege modes must run in either CLIC mode or all privilege
...
-----
-
-
-
-
-
-
+==== smclicshv and the Next Interrupt Handler Address and Interrupt-Enable CSRs ({nxti})
 
+With the smclicshv extension, if the highest ranked interrupt is SHV, {nxti} returns a pointer to the hardware-vectored trap-handler entry.
```
christian-herber-nxp, line 1883: you need to specify what that pointer is. Is this the same logic as for the non-SHV ones?

clic.adoc
```diff
@@ -1878,55 +1878,15 @@ Note: For now all privilege modes must run in either CLIC mode or all privilege
...
-
+==== smclicshv and the Next Interrupt Handler Address and Interrupt-Enable CSRs ({nxti})
 
+With the smclicshv extension, if the highest ranked interrupt is SHV, {nxti} returns a pointer to the hardware-vectored trap-handler entry. 
 
+NOTE: Hardware-vectored and software handlers may have different software interfaces. 
+The assumption is that hardware vectoring would have customized context save/restore, whereas the software vectoring would use a generic context save/restore.
+For example, after saving general context, the software handler could keep processing pending interrupts in a pending-interrupt service loop using {nxti} to claim pending interrupts.
+If the highest ranked interrupt is SHV, this could result in performing unnecessary customized context save/restore if the hardware-vectored handler always saves/restores context. 
+Software implementations may choose to have hardware-vectored handlers check {cause}.{pil} to determine if the customized context save/restore is necessary.
```
christian-herber-nxp, line 1889: I think another handler example might be needed, or extending the existing C-ABI Trampoline Code

christian-herber-nxp, line 1889: Considering only SHV interrupts, this change would allow chaining of vectored interrupts if all handlers conventionally have the same prologue/epilogue, right?

christian-herber-nxp: Thinking about this, I do not think there are many scenarios in which you want to check xcause.pil. Checking will increase your maximum interrupts latency, which is the key parameter for a HW-vectored interrupts. If you are worrying about throughput, the SW-vectored interrupts will be better anyway.

PR #392, opened 2024-03-12
add recommended layout if mem-mapped access to CLIC regs is supported

dansmathers:
As issue #349 discusses, the memory mapped CLIC spec is already implemented in devices.  This pull adds back the memory mapped access details as a recommendation if memory mapped access is provided in addition to indirect CSR access.

### DIFF
--- a/clic.adoc
+++ b/clic.adoc
@@ -3133,6 +3133,128 @@ ID  Interrupt
 2+  other
 ----
 
+== CLIC register access via Memory-Mapped Registers Recommended Format
+If memory-mapped access is provided in addition to access via indirect CSR access, the following regions and access rules are recommended.
+
+=== M-Mode CLIC Register Memory Map Region
+
+Each hart has a separate CLIC accessed by a separate address region.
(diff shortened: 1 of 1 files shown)


clic.adoc
```diff
@@ -3133,6 +3133,128 @@ ID  Interrupt
 2+  other
 ----
 
+== CLIC register access via Memory-Mapped Registers Recommended Format
+If memory-mapped access is provided in addition to access via indirect CSR access, the following regions and access rules are recommended.
```
tovine, line 3137: > [...] in addition to access via indirect CSR

Seems like indirect CSR support is still mandatory, meaning that existing (memory-mapped-only) implementations are technically not compliant.
What's the best course of action here? In my head it doesn't make a lot of sense for an implementation to offer _both_ access methods, so maybe have it _either or_, and define two separate extensions e.g. `Smclicmem` or `Smclicind`?

jb-brelot-nxp:
issue solve. Pull request no more needed.

PR #408, opened 2024-09-11
Improved clarity regarding xcause when switching between CLIC and CLINT

### LINKED ISSUE #405: Conflicting statements about switching from CLIC to CLINT
When you switch from CLIC to CLINT mode the spec says:

>  The other new CLIC xcause fields, xpp and xpie, appear as zero in the xcause CSR but the corresponding state bits in the mstatus register are not cleared.

This doesn't make any sense because it also says:

> The mcause.mpp and mcause.mpie fields typically mirror the mstatus.mpp and mstatus.mpie fields, and are aliased into mcause to reduce context save/restore code.

If they are aliased how do they have different values in `xcause` and `xstatus`?

Also `typically` is very imprecise.

christian-herber-nxp:


jb-brelot-nxp:
Hi

I have discussed with Christian and I will approuve this pull request and then he will come with some fix for the second point.

PR #431, opened 2024-11-11
414 move register layout definitions into wavedrom

wmat:
I've added several register definitions as diagrams and removed the text based descriptions.  If additional register diagrams are required, please open a new issue and assign to me.

### DIFF
--- a/src/clic.adoc
+++ b/src/clic.adoc
@@ -555,7 +555,7 @@ only certain mode/trigger/shv types.
   0       reserved for smclicshv extension (WARL 0)
 ----
 .clicintattr[i] register layout
-include::images/wavedrom/clicintattr.edn[]
+include::images/wavedrom/clicintattri.edn[]
 
 The 2-bit `trig` WARL field specifies the trigger type and polarity for each
 interrupt input. Bit 1, `trig[0]`, is defined as "edge-triggered"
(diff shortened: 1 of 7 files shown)


jb-brelot-nxp:
Hi Bill,

It seems that we can t see properly the result of this pulll request since the build is failling. Could you fix it? thanks

wmat:
docs-resources has been fixed and the PDF for the PR can befound in the Actions tab.

xiaoweish:
With Microsoft Edge version 130.0.2849.80 and PDF-Xchange Editor, both show as below
image


Here is previous version:
image

jb-brelot-nxp:
@wmat it seems that now it is compiling and producing the PDF, but the content is not correct. Something is going wrong.

wmat:
Interesting.  This only seems to be occurring with the CLIC spec.  I'm looking into it.

wmat:
Was there ever a PDF build of this spec?  I ask because I can't get it to build properly using the PDF theme from docs-resources.  The attributes are completely ignored.

wmat:
NVM, I figured out the problem.  Fixing in docs-resources.

wmat:
OK the build artifacts have a fixed PDF now for review.

Issue #434, opened 2024-11-29
Conflicting statements about xnxti write side effects


Timmmm:
In the text:

> Following the usual convention for CSR instructions, if the CSR instruction does not include write side effects (e.g., csrr t0, mnxti), then no state update on any CSR occurs.

But in the code:

```
if (rs1[4:0] != 0 && rs1 != x0) {  // Side-effects should occur.
```

These conflict because "the usual convention for CSR instructions" is *NOT* `rs1[4:0] != 0 && rs1 != x0`, it's just `rs1 != x0`.

For example 

```
li x1, 0
csrrw x2, xnxti, x1
```

is considered to have write side effects by the normal convention, but apparently not by the pseudocode above.

Issue #433, opened 2024-11-14
MNXTI CSR clarifications?


mark-honman:
Section 3.8.7 "Next Interrupt Handler Address and Interrupt-Enable CSR"

It seems like the use of CSRRS was added after the text was written, leading to some inconsistencies. 

- text refers to "mstatus CSR is used for the RMW portion of the operation" but the CSRRS example only affects the bottom 5 bits of mstatus. 
- text refers to "must have a level greater than the saved interrupt level mcause.mpil" but the CSRRS example does not check this value, and instead uses rs1[23:16]. Maybe this is intentional but there is no explanation...
- It seems weird that CSRRS is permitted but not CSRRC.
- Similarly weird, the CSRRSI immediate cannot include non-zero immediate values for bits 0, 2, and 4 but no such restriction applies to CSRRS or CSRRCI. 

Perhaps it would be helpful to be more explicit, say "mie and sie bits of the mstatus CSR are used for the RMW portion of the operation; no other bits may be set in the CSRRS/I or CSRRC/I operand"

It may be useful to describe in the conditions for a "suitable" pending interrupt, that those tests are independent of global interrupt enable (which appears to be the case, by inference from the example interrupt handler code in the SW section).

hirooih:
> - text refers to "must have a level greater than the saved interrupt level mcause.mpil" but the CSRRS example does not check this value, and instead uses rs1[23:16]. Maybe this is intentional but there is no explanation...

I also would like to know the use-case of `csrrs rd, xnxti, rs1 (rs1 != x0)`.  All of sample codes in Appendix use only `csrrsi/csrrci xnxti`.

The use-case for `csrrs rd, xnxti, x0` is described:

> This can be used to determine if an interrupt could be taken without actually updating mil and exccode.

But in this case `mcause.mpil` should be used instead of `x0[23:16]`.  Pseudo-code for this is very simple and easy to implement by reusing logic for `csrrsi/csrrci xnxti`.

```cpp
 if (clic.priv==M && clic.level > mcause.mpil && clic.level > mintthresh.th) {
   rd = VTBASE + XLEN/8 * clic.id; // Return pointer to trap handler entry.
 } else {
   rd = 0;
 }
```

The current specification of `csrrs rd, xnxti, rs1 (rs1 may be x0)` has no known use-case and is not easy to implement.

christian-herber-nxp:
nxti was removed

Issue #415, opened 2024-09-18
Write side-effects for csrrs rd, mnxti, rs1 when rs1 is not x0


MarkHillCodasip:
The spec says it follows the usual CSR conventions for write side-effects when accessing xnxti:

https://github.com/riscv/riscv-fast-interrupt/blob/12f72cfbd92dd973f4696bd03d919a98db0c9ab3/src/clic.adoc?plain=1#L979

The "usual CSR conventions" are  defined in the ISA Volume 1 Spec which says that write-side effects occur whenever rs1 is not x0 even if the register contents is zero:

"Note that if rs1 specifies a register other than x0, and that register holds a zero value, the instruction will not action any
attendant per-field side effects, but will action any side effects caused by writing to the entire CSR."

In particular, when following the usual CSR conventions the occurrence of a write-side effect can be determined purely by decoding the opcode and no read of the source operands is needed to detect this.

However this seems to be contradicted by the pseudo code that includes a check for rs1[4:0] being non-zero:

https://github.com/riscv/riscv-fast-interrupt/blob/12f72cfbd92dd973f4696bd03d919a98db0c9ab3/src/clic.adoc?plain=1#L1029

I'd suggest reducing this line just to "if (rs1 != x0)"

jb-brelot-nxp:
need to be updated in two places.

WIl lcare about this week.

christian-herber-nxp:
feature was removed

Issue #399, opened 2024-07-15
an exception trap on the trap handler function address fetch cannot return to the instruction interrupted (smclicshv)


hirooih:
8.2.1. smclicshv Changes to xtvec CSR Mode for CLIC

> If the trap handler function address fetch is unsuccessful and a exception trap occurs, the `xinhv` bit in `xcause` of the exception handler privilege mode is set indicating that `xepc` of the exception handler privilege mode contains a trap handler function address instead of the virtual address of an instruction.

I thinks this will lose the virtual address of the instruction interrupted.

The last note on this section writes:

> Interrupted context is lost on horizontal traps during table fetch where exceptions are the same privilege level as the interrupt handler.

But vVirtual traps also lose the interrupted context.

zcmt has similar issue.

Unprivileged Architecture, Version 20240411, 27.14.2. Table Jump Fault handling

> If an exception occurs on either instruction fetch, xEPC is set to the PC of the table jump instruction, xCAUSE is set as expected for the type of fault and xTVAL (if not set to zero) contains the fetch address which caused the fault.

I think a similar simple approach would work well.

-  Set `xepc` to the address of next instruction interrupted as usual interrupt. 
-  Set `xtval` to the addreess of the trap handler function address.
- After `xret` from load access-fault or load page-fault (misaligned exception should not occurs, because trap vector table is aligned.) handler, the interrupt should occurs again.
- In this method `xinhv` bit is not required.

I'm afraid that I'm missing something, but please let me know.

christian-herber-nxp:
inhv removed https://github.com/riscv/riscv-fast-interrupt/pull/513

PR #438, opened 2025-01-05
fix inconsistencies on xnxti CSR

hirooih:
Regarding the definition of xnxti CSR, several inconsistencies have been pointed out in #395, #415, #433, and #434. This PR attempts to resolve them.

***

In addition to this PR, additional explanations are needed for the following points.

- Why the csrrs instruction compares clic.level with rs1[23:16] instead of mcause.mpil.
- Why the csrrs instruction is permitted but not csrrc instruction.
- Usecases of the csrrs instruction for this CSR.

***

I removed the following comment after each pseudo code.

> When a different CSR instruction is used, the update of mstatus and the test
> for whether side-effects should occur are modified accordingly.

This conflicts with the following sentence. (Using different CSR is reserved.)

> Accessing the {mnxti} CSR using any other CSR instruction (i.e., CSRRW, CSRRC, or CSRRWI) is reserved.

### DIFF
--- a/src/clic.adoc
+++ b/src/clic.adoc
@@ -1004,12 +1004,19 @@ NOTE: Vertical interrupts to higher privilege modes will be taken
 preemptively by the hardware, so {mnxti} effectively only ever handles
 the next interrupt in the same privilege mode.
 
-Pseudo-code for csrrsi rd, mnxti, uimm[4:0] in M mode:
+Pseudo-code for csrrsi|csrrci rd, mnxti, uimm[4:0] in M mode:
 [source]
 ----
  // clic.priv, clic.level, clic.id represent the highest-ranked
  // interrupt currently present in the CLIC
- mstatus |= uimm[4:0]; // Performed regardless of interrupt readiness.
+ // Performed regardless of interrupt readiness.
+ if (funct3 == csrrsi) {
+     mstatus.mie |= uimm[3];
+     mstatus.sie |= uimm[1];
+ } else {   // csrrci
+     mstatus.mie &= !uimm[3];
+     mstatus.sie &= !uimm[1];
+ }
  if (clic.priv==M && clic.level > mcause.mpil && clic.level > mintthresh.th) {
    // There is an available interrupt.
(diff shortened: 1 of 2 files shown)


src/clic.adoc
```diff
@@ -1037,13 +1042,12 @@ Pseudo-code for csrrs rd, mnxti, rs1 in M mode:
...
-   {
-      mstatus |= rs1[4:0]; // Performed regardless of interrupt readiness.
-   }
+   // Performed regardless of interrupt readiness.
+   mstatus.mie |= rs1[3];
+   mstatus.sie |= rs1[1];
    if (clic.priv==M && clic.level > rs1[23:16] && clic.level > mintthresh.th) {
      // There is an available interrupt.
-     if (rs1[4:0] != 0 && rs1 != x0) {  // Side-effects should occur.
+     if (rs1 != x0) {  // Side-effects should occur.
```
jb-brelot-nxp, line 1050: This changes can t be accepted since it is changing the initial behavior.

Currently I m starting a thread of mail with ARC to better understand what we should do.

hirooih: Thank you for your response.

> This changes can t be accepted since it is changing the initial behavior.

This PR does not propose a change to the specifications.

We have to change either the pseudo code or the following descriptions.

> If the CSR instruction that accesses mnxti includes a write, ...

> Following the usual convention for CSR instructions, if the CSR instruction does not include write side effects (e.g., csrr t0, mnxti), then no state update on any CSR occurs.

> Currently I m starting a thread of mail with ARC to better understand what we should do.

Thank you!

***

On the other hand, #399, I opened last July, requires changing the behavior. The current specification simply does not work (If I understand correctly).

jb-brelot-nxp: A thread of mail has been started to understand the best way to proceed with NXTI. 
I ll keep you updated

hirooih: @jb-brelot-nxp 

I've updated this PR by reading the discussion on the ML.

christian-herber-nxp:
nxti was removed

PR #454, opened 2025-03-04
Updating xcliccfg register definitions

### LINKED ISSUE #453: align definition of clifcfg CSR with how its done in the priv manual
there is currently too much text in the register definition. this makes it difficult to add more fields without overruning bounding boxes.

christian-herber-nxp:


christian-herber-nxp:
@wmat : I could not figure out how to reduce the size of a field without breaking the build. Can you check this, as currently then spans are too large after removing a bunch of the text.

smatherd:
ok to close for now since cliccfg isn't in the current version of the spec?

PR #458, opened 2025-03-05
Defined the first CLIC interrupts to be an alias of mie/mip

### LINKED ISSUE #456: Undo the removal of xie / xip
Since the change to indirect CSRs, these can have a comeback as an alias of the indirect registers.

christian-herber-nxp:


tovine (approved):
I like these changes, they simplify the interaction with basic interrupts and removes a lot of the previous uncertainty and "platform defined behavior" regarding the lower interrupt indices.
But does this also mean that all external IRQs essentially have to start at CLIC index 1, or is there an offset between external IRQ number and `xcause` IRQ number?

Note that I've been out of the loop for a bit (busy with other tasks) so I might be missing something that's obvious in the spec, please forgive that ignorance.

src/clic.adoc
```diff
@@ -501,6 +497,11 @@ Each interrupt input has a dedicated interrupt pending bit
 Software should assume `clicintip[__i__]=0` means no interrupt pending, and
 `clicintip[__i__]=1` indicates an interrupt is pending.
 
+The the lowest index `clicintip[__i__]` bits are an alias of the `mip` CSR.
```
Silabs-ArjanB, line 500: The the -> The

src/clic.adoc
```diff
@@ -544,12 +544,16 @@ the global enable bit for the current privilege mode. Therefore, for an
 interrupt `_i_` to be enabled in the current privilege mode, both `clicintie[__i__]`
 and {status}.{ie} have to be set.
 
-
 NOTE: In contrast, since {status}.{ie} only takes effect in the current privilege
 mode according to RISC-V convention, an interrupt `_i_` from a higher privilege mode
 is enabled as long as `clicintie[__i__]` is set (regardless of the setting
 of {status}.{ie} in the higher privilege modes).
 
+The the lowest index `clicintie[__i__]` bits are an alias of the `mie` CSR.
```
Silabs-ArjanB, line 552: The the -> The

PR #459, opened 2025-03-05
Allowing level-based interrupts to have a SW-settable bit

### LINKED ISSUE #450: Value of clicintip when switching to edge-triggered mode should be defined
> The value in the clicintip[i] is undefined when switching from level-sensitive mode to edge-triggered mode in clicintattr[i].

This seems unnecessarily timid. Is there a good reason not to zero it?

### LINKED ISSUE #457: Possible conflict with behavior Machine mode for behavior of pending bits
Machine mode states this:

Supervisorlevel-external interrupts are made pending based on the logical-OR of the software-writable SEIP bit
and the signal from the external interrupt controller.

While the CLIC spec requires the pending bit to always reflect the external wire state if they are level based interrupts. The mentioned behavior is only possible for edge triggered interrupts, but I am not sure we would be able to enforce this (or could we?)

christian-herber-nxp:


src/clic.adoc
```diff
@@ -505,34 +505,36 @@ The conditions for an interrupt trap to occur must be evaluated in a bounded amo
...
-by hardware interrupt inputs and by software.  The bit is set by
-hardware after an edge of the appropriate polarity is observed on the
-interrupt input, as determined by the `clicintattr[__i__]` field.
-Software writes to `clicintip[__i__]` can set or
-clear edge-triggered pending bits directly by writes to the
+Note: As per the machine-level ISA, this behavior is required for the SEI interrupt (exccode 9).
 
-`clicintip[__i__]` register. Edge-triggered pending bits can also be cleared when a CSR instruction that accesses {nxti} includes a write.
-`clicintip[__i__]` behavior is unaffected by `clicintie[__i__]` setting.
+If `clicintip[__i__]` is writable, a pending interrupt i can be cleared by writing 0 to this bit.
```
Timmmm, line 526: If the input wire is still 1 it wouldn't be cleared though right? Since the software bit is ORed with it?

src/clic.adoc
```diff
@@ -505,34 +505,36 @@ The conditions for an interrupt trap to occur must be evaluated in a bounded amo
...
+a writable pending bit for software vectored interrupts.  Additional detail
 on this is described in the {nxti} section.
 
-The value in the `clicintip[__i__]` is undefined when switching from
-level-sensitive mode to edge-triggered mode in `clicintattr[__i__]`.
-
-NOTE: Software cannot rely on the underlying `clicintip[__i__]`
-register bits used in edge-triggered mode to hold state while in
-level-sensitive mode.
+The value in the `clicintip[__i__]` is undefined
```
Timmmm, line 535: Did you mean to remove this line?

src/clic.adoc
```diff
@@ -505,34 +505,36 @@ The conditions for an interrupt trap to occur must be evaluated in a bounded amo
...
+When switching from level-sensitive mode to edge-triggered mode,
+`clicintattr[__i__]` has the value of the software-writable bit, ignoring the previous external signal level.
```
Timmmm, line 537: What if it isn't writable in level-triggered mode? Presumably it becomes 0?

Timmmm (commented):
I always found this super confusing with SEI since it's basically using the same name for two different signals.

Could you introduce a new names for the input wire values and maybe the writable software bit so that it's a bit easier to talk about? Then you can also avoid saying "When writable" everywhere and instead just say "the writable software bit can be read-only zero" (I'm sure you can come up with a better name than "the writable soft bit"!)

Timmmm (commented):
I always found this super confusing with SEI since it's basically using the same name for two different signals.

Could you introduce a new names for the input wire values and maybe the writable software bit so that it's a bit easier to talk about? Then you can also avoid saying "When writable" everywhere and instead just say "the writable software bit can be read-only zero" (I'm sure you can come up with a better name than "the writable soft bit"!)

christian-herber-nxp:
Quite explicitly, it is still an implementation choice for a level based interrupt to not have a writable pending bit. Otherwise you are spending a lot of flip flips where they are not needed.

Timmmm:
> Quite explicitly, it is still an implementation choice for a level based interrupt to not have a writable pending bit. Otherwise you are spending a lot of flip flips where they are not needed.

Right but conceptually you can describe it as having one for simplicity. Then you also say "it can be read-only zero", which would allow implementations where it doesn't actually exist.

src/clic.adoc
Silabs-ArjanB: Agreed with @Timmmm. Shouldn't this state that interrupt clearing might also need to be performed at the very source (driving the interrupt wire)?

src/clic.adoc
```diff
@@ -965,7 +967,7 @@ If the CSR instruction that accesses {mnxti} includes a write, the
 {mstatus} CSR is used for the read-modify-write portion of the
 operation, while the {mcause} register's `exccode` field and the
 {mintstatus} register's {mil} field can also be updated with the new interrupt id and level.
-If the interrupt is edge-triggered, then the pending bit is also zeroed.
+If the interrupt is writable, then the pending bit is also zeroed.
```
Silabs-ArjanB, line 970: is writable -> has a writable pending bit

Silabs-ArjanB (changes requested):

src/clic.adoc
smatherd: To be consistent with AIA spec, I think we should stay out of the business of what happens when switching from level-sensitive to edge-triggered.  see issue #450.

PR #460, opened 2025-03-08
attempt at creating a AIA clic extension

smatherd:
Tried to separate CLIC spec into feature categories.  AIA already has a priority method so I removed the ability of clicintctl to represent a mix of level and priority.  clicintctl was renamed clicintlvl.  I also moved to the ideleg method of assigning privilege modes to interrupts.  With these changes, CLIC parameters was simplified and smclicconfig was rendered unnecessary.  Since clicintlvl only impacts the current privilege mode and CSRs mip/mie are mirrors of the lower 64 interrupts, there no longer appears that there is a need to require all priv modes are CLIC or all priv modes are CLINT.

### DIFF
--- a/src/aia_clic_extension.adoc
+++ b/src/aia_clic_extension.adoc
@@ -170,7 +170,6 @@ This table provides a summary of the CLIC extensions to AIA.
 | smclicconfig | Allows implementations to support different parameterizations of CLIC extensions
 | smclicdbg    | support for interrupt debug triggering
 | smclicip     | support for interrupt context push/pop
-| smcliclat    | features to improve latency of context switching.
 |===
 
 == Increase of AIA local interrupts - smclicincr
@@ -565,7 +564,6 @@ interrupt's privilege mode (`**__x__**`) and interrupt level (`il`).
 A new M-mode CSR, `mpintstatus`, holds consolidated state of preempted context.
 
 NOTE: Previous versions of CLIC modified the {cause} CSR to contain preempted context.
-That modification has been moved to the smcliclat extension.
 {pintstatus} has been added to allow implementations to maintain {cause} compatibility between
 CLIC and CLINT modes.
 
--- a/src/aia_clic_extension.adoc
+++ b/src/aia_clic_extension.adoc
@@ -164,7 +164,8 @@ This table provides a summary of the CLIC extensions to AIA.
 | Extension Name | Description
 | smclicincr   | Increase of the number of local interrupts for M-mode
 | ssclicincr   | Increase of the number of local interrupts for S-mode
-| smclic       | Same Privilege Mode Interrupt Preemption support for M-mode
+| smclic       | Horizontal Nested Interrupt Preemption support for M-mode
+
 | ssclic       | Same Privilege Mode Interrupt Preemption support for S-mode
 | smclicshv    | Selective Hardware Vectored Interrupts
 | smclicconfig | Allows implementations to support different parameterizations of CLIC extensions
--- a/src/aia_clic_extension.adoc
+++ b/src/aia_clic_extension.adoc
@@ -165,8 +165,7 @@ This table provides a summary of the CLIC extensions to AIA.
 | smclicincr   | Increase of the number of local interrupts for M-mode
 | ssclicincr   | Increase of the number of local interrupts for S-mode
 | smclic       | Horizontal Nested Interrupt Preemption support for M-mode
-
-| ssclic       | Same Privilege Mode Interrupt Preemption support for S-mode
+| ssclic       | Horizontal Nested Interrupt Preemption support for S-mode
 | smclicshv    | Selective Hardware Vectored Interrupts
 | smclicconfig | Allows implementations to support different parameterizations of CLIC extensions
 | smclicdbg    | support for interrupt debug triggering
@@ -190,9 +189,6 @@ NOTE: The existing timer (`mtip`/`stip`), software
 sources, where the privilege mode, interrupt level, and priority can
 be altered using `clicintattr[__i__]` and
 `cliciprio[__i__]` and `clicmideleg[__i__]` registers.
-
-NOTE: To maintain mip behavior, interrupts with read-only behavior should be configured/hard-coded with trigger type level and
-interrupts that are writable should be configured/hard-coded with trigger type edge and not driven by a hardware interrupt input. 
 
 With this extension, for implementations that do not require CLINT compatibility, the allocation of interrupt ordering (e.g. meip, mtip, msip) 
 are allowed to be defined by the platform. A platform definition will often be based on a specific RISC-V ISA profile, 
@@ -275,9 +271,9 @@ of {status}.{ie} in the higher privilege modes).
 
 NOTE: This register bit is defined as WARL as unimplemented interrupts appear hardwired to zero.
 
-==== Interrupt Input Identification Number
+==== Major interrupt numbers
 
-The 4096 CLIC interrupt inputs are given unique identification numbers
+CLIC provides up to 4096 major interrupt numbers indicated
 with {mcause} Exception Code (`exccode`) values. 
 
 === Indirect Access M-mode CLIC interrupt CSRs
@@ -301,7 +297,7 @@ although this is expected to be an atypical need for most interrupt handlers.
 
 In this `miselect` offset range:
 
-* Each `mireg2` register controls the clic attribute setting of four interrupts
+* When XLEN = 32, each `mireg2` register controls the clic attribute setting of four interrupts, with one 8-bit byte per interrupt.
 
 [%autowidth]
 |===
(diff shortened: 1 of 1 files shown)


src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+Controller (CLIC) extensions to AIA.
+
+This table provides a summary of the CLIC extensions to AIA.
+
+[%autowidth]
+|===
+| Extension Name | Description
+| smclicincr   | Increase of the number of local interrupts for M-mode
+| ssclicincr   | Increase of the number of local interrupts for S-mode
+| smclic       | Same Privilege Mode Interrupt Preemption support for M-mode
```
christian-herber-nxp, line 167:
```suggestion
| smclic       | Horizontal Nested Interrupt Preemption support for M-mode

```

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+| ssclic       | Same Privilege Mode Interrupt Preemption support for S-mode
+| smclicshv    | Selective Hardware Vectored Interrupts
```
christian-herber-nxp, line 169: General comment. I think we should drop selective HW vectoring. It should be simply hardware vectoring. Making it selective is fairly costly (adding at least one FF per source). We will make HW vectoring so attractive that if you have the ability, you will want all interrupts to be HW vectored.

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+| smclicconfig | Allows implementations to support different parameterizations of CLIC extensions
+| smclicdbg    | support for interrupt debug triggering
+| smclicip     | support for interrupt context push/pop
+| smcliclat    | features to improve latency of context switching.
+|===
```
christian-herber-nxp, line 174: I think we should add another extension for the using handler addresses in the table instead of jumps.
This is a very platform specific choice, and it does not affect C code etc. 
It really only affects platform specific code like startup code and linker file.
Also, I think it would be good to have this consistent between all kinds of exceptions. This would eliminate the need for mtvt register and everything can work with mtvec (which will make CLIC more compatible with the rest)

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+(`clicmideleg[__i__]`) to delegate interrupts to a lower privilege level.
+
+NOTE: The existing timer (`mtip`/`stip`), software
+(`msip`/`ssip`), and external interrupt inputs
+(`meip`/`seip`) can be treated as additional local interrupt
+sources, where the privilege mode, interrupt level, and priority can
+be altered using `clicintattr[__i__]` and
+`cliciprio[__i__]` and `clicmideleg[__i__]` registers.
+
+NOTE: To maintain mip behavior, interrupts with read-only behavior should be configured/hard-coded with trigger type level and
```
christian-herber-nxp, line 193: I am not sure if this is a note or a normative requirement.
Also, requiring the edge triggered behavior might be too restrictive.

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+NOTE: In contrast, since {status}.{ie} only takes effect in the current privilege
+mode according to RISC-V convention, an interrupt `_i_` from a higher privilege mode
+is enabled as long as `clicintie[__i__]` is set (regardless of the setting
+of {status}.{ie} in the higher privilege modes).
+
+NOTE: This register bit is defined as WARL as unimplemented interrupts appear hardwired to zero.
+
+==== Interrupt Input Identification Number
+
+The 4096 CLIC interrupt inputs are given unique identification numbers
```
christian-herber-nxp, line 279: This is called a major identity number in AIA terms, we should stick to that

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+
+In this `miselect` offset range:
+
+* Each `mireg2` register controls the clic attribute setting of four interrupts
+
+[%autowidth]
+|===
+| `miselect` |  `mireg2` bits |  `mireg2` state              | description
+
+| 0x1000+i   |   7:0          | RW  `clicintattr[__i__*4+0]` |  setting for interrupt __i__*4+0
```
christian-herber-nxp, line 309: the general strategy of register layout is different in AIA for RV64. When RV64, only every second register is used, and each CSR accesses 64b. This does increase efficiency for RV64

christian-herber-nxp: and when I say RV64, i mean the the XLEN at the current mode is 64 :)

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+
+== Same Privilege Mode Interrupt Preemption Support - smclic
+
+This extension requires smclicincr extension.
+
+The smclic extension extends interrupt preemption to support up to 256 interrupt
+levels for each privilege mode, where higher-numbered interrupt levels
+can preempt lower-numbered interrupt levels.  Interrupt level 0
+corresponds to regular execution outside of an interrupt handler.
+Levels 1--255 correspond to interrupt handler levels. Platform
```
christian-herber-nxp, line 434: the statement on platform levels is not needed here. otherwise you could put this anywhere. Also, I believe an ISA profile would also be allowed to restrict this.

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+privilege mode.
+
+NOTE: Existing RISC-V interrupt behavior is retained, where incoming
+interrupts for a higher privilege mode can preempt an active interrupt
+handler running in a lower privilege mode, regardless of global
+interrupt enable in lower privilege mode.
+
+=== CLIC Interrupt Input Control (`clicintlvl`)
+
+`clicintlvl[__i__]` is an 8-bit WARL control register
```
christian-herber-nxp, line 449: this is an additional register? Why do we not just use some of the bits of iprio?

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+
+NOTE: Selecting an interrupt at a high privilege mode masks any
+interrupt at a lower privilege mode since the higher-privilege mode
+causes the interrupt signal to appear more urgent than any lower-privilege
+mode interrupt.
+
+NOTE: This register is defined as WARL as some implementations may want to support a limited number
+of values in this register including hardwiring some bits to fixed values.
+
+NOTE: Within a single privilege mode, it can be useful to separate interrupt
```
christian-herber-nxp, line 473: I do not see the value add of this note.

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+handler tasks from application tasks to increase robustness, reduce
+space usage, and aid in system debugging.  Interrupt handler tasks
+have non-zero interrupt levels, while application tasks have an
+interrupt level of zero.
+
+=== New {mtvec} CSR Mode for CLIC
```
christian-herber-nxp, line 479: mtvec is not really a preemption feature. As noted above, it should be its own extension.

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+mpintstatus 
+Bits   Field         Description
+ XLEN-1 Interrupt     Set to 1 for interrupts and 0 for exceptions.  Mirror of mcause.Interrupt
+ 30     (reserved)    Contains minhv (AKA xinhv) bit when smclicshv extension present
+ 29:28  mpp[1:0]      Previous privilege mode, mirror of mstatus.mpp
+ 27     mpie          Previous interrupt enable, mirror of mstatus.mpie
+ 26:24  (reserved)
+ 23:16  mpil[7:0]     Previous interrupt level
+ 15:12  (reserved)
+ 11:0   exccode[11:0] Exception/interrupt code, mirror of mcause[11:0]
```
christian-herber-nxp, line 534: I am not sure we need exccode here. We should discuss this separately I believe. If anything, we would want the previous exccode. However, since we are dropping nxti (thats my assumption), and will advertise for push/pop, i think the problem we were facing might go away.

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+----
+
+
+As stated in the RISC-V privilege specification,
```
christian-herber-nxp, line 538: It might be good to avoid restating these things, but to rather specify incrementally. Also, you will want to specify on top of the AIA behavior. There is also other specifications which modify this behavior, e.g. double trap.

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+from CLIC mode causes {pil} in that privilege mode to be zeroed.
+
+
+==== New Interrupt Status ({mintstatus}) CSR
+
+A new M-mode CSR, `mintstatus`, holds the active interrupt level for
+each supported privilege mode.  These fields are read-only.  The
+primary reason to expose these fields is to support debug.
+
+[source]
```
christian-herber-nxp, line 557: In RISC-V, you cannot determine the mode you are in. Does this contradict this?

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+
+Within a privilege mode `*_x_*`, if the associated global
+interrupt-enable {ie} is clear, then no interrupts will be taken in
+that privilege mode, but a pending-enabled interrupt in a higher
+privilege mode will preempt current execution.  If {ie} is set, then
+pending-enabled interrupts at a higher interrupt level in the same
+privilege mode will preempt current execution and run the interrupt
+handler for the higher interrupt level.
+
+As with the existing RISC-V mechanism, when an interrupt or
```
christian-herber-nxp, line 647: Some of this seems to be a repeat of an earlier section

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+ P  1  ?  | nP=P     0      ?  |->    - -  -   -   -     -  -  -  -   # No interrupt
+ P  1  L  | nP=P   0<nL<=L  ?  |->    - -  -   -   -     -  -  -  -   # Interrupt ignored
+ P  1  L  | nP=P   L<nL    id  |->    P 0  nL  V   id    P  L  1  pc  # Horizontal interrupt taken
+ P  ?  ?  | nP>P     0      ?  |->    - -  -   -   -     -  -  -  -   # No interrupt
+ P  e  L  | nP>P   0<nL    id  |->   nP 0  nL  V   id    P  L  e  pc  # Vertical interrupt taken
+----
+
+==== Critical Sections in Interrupt Handlers
+
+To implement a critical section between interrupt handlers at
```
christian-herber-nxp, line 684: this is a repeat of an earlier section

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+.A pending-and-enabled interrupt _i_ causes the hart to resume execution if interrupt _i_
+* has the same privilege mode as the current privilege mode and
+* the interrupt priority reduction tree selects interrupt _i_ as the maximum across all pending-and-enabled interrupts and
+* the interrupt _i_ level is greater than max({intstatus}.{il}, {intthresh}.`th` )
+
+.A pending-and-enabled interrupt _i_ causes the hart to resume execution if interrupt _i_
+* has a lower privilege mode than the current privilege mode and
+* the interrupt priority reduction tree selects interrupt _i_ as the maximum across all pending-and-enabled interrupts and
+
+NOTE: {intthresh} only applies to the current privilege mode.  There is a proposal to add a new WFMI instruction ("wait for mode's interrupts") to the privilege specification. This instruction only has to wakeup for pending-and-enabled interrupts in the current mode, and is not required to wakeup for pending-and-enabled interrupts in lower privilege modes. Pending-enabled higher privilege-mode interrupts will interrupt/wakeup as usual.
```
christian-herber-nxp, line 722: where is this proposal? If this did not materialize and is no longer wanted, the note should be removed

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+
+Horizontal synchronous exception traps, which stay within a privilege
+mode, are serviced with the same interrupt level as the instruction
+that raised the exception.
+
+Vertical synchronous exception traps, which are serviced at a higher
+privilege mode, are taken at interrupt level 0 in the higher privilege
+mode.
+
+WARNING: Traps should be avoided at any time when {epc}/{cause} are live
```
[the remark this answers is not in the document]
christian-herber-nxp: also, i think this warning should refer explicitly to Synchronous exception traps

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+because these CSRs will be overwritten. Software should try to back them
+up if needed.
+
+==== Non-Resumable Non-Maskable Interrupts
+
+The handling of NMIs is implementation-specific, but NMIs are always
```
christian-herber-nxp, line 740: I feel this doesn't deserve a section. That is just the definition of NMI and nothing is new with respect to CLIC

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+handled in M-mode and can overwrite `mepc` and `mcause` of an active
+M-mode CLIC interrupt handler.
+
+==== Resumable Non-Maskable Interrupts
```
christian-herber-nxp, line 744: Similarly, I do not know why we have to state that it is orthogonal. 
You do not want to add a section for every extension that is orthogonal.

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+NOTE: This section describes the interaction of the CLIC with the
+proposed new RNMI specification.
+
+The resumable NMI (RNMI) extension adds additional `mnepc`, `mncause`,
+and `mnstatus` CSRs.  The RNMI CSRs are separate from the CLIC CSRs and so CLIC state does not need to be saved in RNMI CSRs.
+This makes the design of the RNMI extension orthogonal to the hart's general interrupt scheme.
+
+==== Returns from Handlers
+
+As stated in the RISC-V Privilege Specification,
```
christian-herber-nxp, line 755: Here also, do not repeat the privileged specification (what does that even mean? - there is only extension, so it would be the Machine or Supervisor Mode ISAs), but specify the changes.

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+Additionally in CLIC mode, {ret} sets {intstatus}.{il} to {pintstatus}.{pil}.
+The {ret} instruction does not modify the {pintstatus}.{pil} field in {pintstatus}.
+
+
+=== CLIC Level Control - Interrupts at machine level
+
+[source]
+----
+       Number  Name         Description
+ (NEW) 0x346   mpintstatus  Previous interrupt levels
```
christian-herber-nxp, line 768: this is more than levels. should be interrupted context or something (if we remove exccode)

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+|===
+| `miselect` |  `mireg` bits |  `mireg` state              | description
+
+| 0x1000+i   |  7:0          | RW  `clicintlvl[__i__*4+0]` |  setting for interrupt __i__*4+0
+| 0x1000+i   | 15:8          | RW  `clicintlvl[__i__*4+1]` |  setting for interrupt __i__*4+1
+| 0x1000+i   | 23:16         | RW  `clicintlvl[__i__*4+2]` |  setting for interrupt __i__*4+2
+| 0x1000+i   | 31:24         | RW  `clicintlvl[__i__*4+3]` |  setting for interrupt __i__*4+3
+|===
+
+=== Interrupts at supervisor level
```
christian-herber-nxp, line 796: this should go under a ssclic heading

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+For implicit hardware vector table fetches, whether breakpoints trap
+on the table read is left as an implementation option. For explicit
+loads used in software vectoring, watchpoints operate as normal for
+any load.  In CLIC mode, the `dpc` CSR additionally may hold the
+faulting address if breakpoints are allowed to trap on the table fetch
+during hardware vectoring.  If breakpoints are allowed to trap on the
+table read, dret should honor {inhv}.
+
+
+== Interrupt trigger Debug extension- smclicdbg
```
christian-herber-nxp, line 1102: I would suggest an extension name without CLIC, as I believe this could easily be implemented on non-CLIC systems.

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+CLICMAXID      12-4095      Largest interrupt ID
+
+INTTHRESHBITS  1-8          Number of bits implemented in {intthresh}.th
+CLICMTVECALIGN >= 2         Number of hardwired-zero LSBs in mtvec address.
+----
+NOTE: These parameters are likely to be available by the general
+discovery mechanism that is in development.
+
+
+== Support for reduced csr accesses in interrupt handlers - smcliclat
```
christian-herber-nxp, line 1446: I think it would be good to remove this. We can always bring it back if needed, but this would increase clarity of the current proposal.

src/aia_clic_extension.adoc
christian-herber-nxp: In fact, I believe this should be a Sd* extension.

src/aia_clic_extension.adoc
smatherd: I think if an implementation wanted to have all interrupts hardware vector since the attr is WARL, it is allowed to tie them all to high.   So that is just a subset of what selective HW vectoring allows.

src/aia_clic_extension.adoc
smatherd: so the ability to have horizontal preemption (intstatus, etc.) when mtvec == 01?

src/aia_clic_extension.adoc
smatherd: I think trying to combine levels and priority in the same register and the flexibility to adjust which bits are level and which are priority is unnecessary FW ABI complexity.  implementations can be clever and provide that flexibility under the hood but keeping the levels separate from priority, especially when lower AIA priority values are higher priority but higher level values  are allow to preempt.  So the proposal is to rename clicintctl as clicintlvl and extend AIA iprio to more interrupts.

christian-herber-nxp: as per my comment below, i think we should also have lower levels take priority, which will simplify the selection algorith,

src/aia_clic_extension.adoc
christian-herber-nxp: according to that argument, shv does not have to be an extension

christian-herber-nxp: TG meeting: consensus is that a mix of HW and SW vectoring is unusual. We should attempt to define a way to have HW-vectoring without shv.

src/aia_clic_extension.adoc
```diff
@@ -376,8 +419,8 @@ locations appear hardwired to zero.
 
 In this `siselect` offset range:
 
-* Each `sireg` register controls the clic level/priority setting of four interrupts
-* Each `sireg2` register controls the clic attribute setting of four interrupts
+* When XLEN = 32, each `sireg` register controls the clic level/priority setting of four interrupts
```
christian-herber-nxp, line 422: I believe we need to be more precise than XLEN. E.g. at M-mode, it would MXLEN, which can be different from XLEN.
I.e. XLEN at the current mode or something.

christian-herber-nxp:
@dansmathers : I would suggest to merge this soon and create issues for the pending topics, then we can work in parallel. I can help out on the work, but I cannot while this PR is pending.
I would just like if we replace the clic document instead of adding a second one before merging to main

src/aia_clic_extension.adoc
smatherd: probably better to stay consistent with AIA.  I see this in the AIA spec:

>Note that the widths of siselect and sireg are always the current XLEN rather than SXLEN.
Hence, for example, if MXLEN = 64 and SXLEN = 32, then these registers are 64 bits when the
current privilege mode is M (running RV64 code) but 32 bits when the privilege mode is S (RV32
code).

I think keeping it XLEN is consistent with AIA, right?

src/aia_clic_extension.adoc
christian-herber-nxp: https://github.com/riscv/riscv-fast-interrupt/issues/465

src/aia_clic_extension.adoc
christian-herber-nxp: https://github.com/riscv/riscv-fast-interrupt/issues/466

src/aia_clic_extension.adoc
christian-herber-nxp: https://github.com/riscv/riscv-fast-interrupt/issues/467

src/aia_clic_extension.adoc
```diff
@@ -0,0 +1,1816 @@
...
+The reset behavior of other fields is platform-specific.
+
+=== smclic Interrupt Operation
+
+This section describes the operation of smclic interrupts.
+
+==== General Interrupt Overview
+
+At any time, a hart is running in some privilege mode with some
+interrupt level.  The hart's privilege mode is held internally but is not visible to software running on a hart (to avoid
```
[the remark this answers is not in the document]
christian-herber-nxp: https://github.com/riscv/riscv-fast-interrupt/issues/469

src/aia_clic_extension.adoc
christian-herber-nxp: https://github.com/riscv/riscv-fast-interrupt/issues/470

src/aia_clic_extension.adoc
christian-herber-nxp: https://github.com/riscv/riscv-fast-interrupt/issues/471

src/aia_clic_extension.adoc
christian-herber-nxp: https://github.com/riscv/riscv-fast-interrupt/issues/469

src/aia_clic_extension.adoc
christian-herber-nxp: https://github.com/riscv/riscv-fast-interrupt/issues/473

src/aia_clic_extension.adoc
christian-herber-nxp: https://github.com/riscv/riscv-fast-interrupt/issues/473

PR #477, opened 2025-04-01
For issue #475 - rework section headers

smatherd:
Moved around sections to better match AIA format.  Instead of having separate smclic/ssclic sections, match AIA where you describe the extension behavior and then have subsections describing how things are done at machine level and then at supervisor level.


christian-herber-nxp:
this is conflicting with the master branch. Changes need to be rebased on top of the new document.

PR #478, opened 2025-04-03
replacing pull #477 - rework section headers

smatherd:
Moved around sections to better match AIA format. Instead of having separate smclic/ssclic sections, match AIA where you describe the extension behavior and then have subsections describing how things are done at machine level and then at supervisor level.

### DIFF
--- a/src/clic.adoc
+++ b/src/clic.adoc
@@ -80,6 +80,7 @@ Graphics used are either explicitly available for free, are property of RISC-V I
 :mtval: pass:q[``mtval``]
 :mip: pass:q[``mip``]
 :mnxti: pass:q[``mnxti``]
+:mpintstatus: pass:q[``mpintstatus``]
 :mintstatus: pass:q[``mintstatus``]
 :mintthresh: pass:q[``mintthresh``]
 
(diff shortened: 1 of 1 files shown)


smatherd:
closed with #478

src/clic.adoc
```diff
@@ -1226,6 +1236,19 @@ Else, the xcspw instruction can be used to atomically swap the stack pointer wit
 This mechanism can also be used to initialize the sp value of a less privileged mode.
 If an operating system manages separate stacks per software thread in a lower privileged mode, this mechanism can also be applied.
 
+=== Smcspsw Conditional Stack Pointer Swap at supervisor level
```
christian-herber-nxp, line 1239: supervisor extensions start with Ss...

christian-herber-nxp:
> Moved around sections to better match AIA format. Instead of having separate smclic/ssclic sections, match AIA where you describe the extension behavior and then have subsections describing how things are done at machine level and then at supervisor level.

I am not sure if AIA is the right role model.
Splitting things clearly by extension per privilege level is the way the privileged manual is handling things. 
I believe the Cheri TG was just asked to rework there spec to have this clear split.
The profiles had to define Ssaia and Smaia extensions cause this definition is missing in AIA.

PR #481, opened 2025-04-03
Removing text specifying ret behavior with register definitions

### LINKED ISSUE #474: Return from handler should not be repeated
The CLIC spec repeats the return from handler behavior.

> As stated in the RISC-V Privilege Specification,
> when executing an xRET instruction, supposing xPP holds the value y, xIE is set to xPIE; the
> privilege mode is changed to y; xPIE is set to 1; and xPP is set to the least-privileged supported
> mode.  The {ret} instruction sets the pc to the value stored in the {epc} register.
> Additionally in CLIC mode, {ret} sets {intstatus}.{il} to {pintstatus}.{pil}.
> The {ret} instruction does not modify the {pintstatus}.{pil} field in {pintstatus}.

It should only specify the changes to the baseline behavior, otherwise it can become incompatible with other changes to that behavior.

christian-herber-nxp:
- there is a dedicated section on the ret behavior

### DIFF
--- a/src/clic.adoc
+++ b/src/clic.adoc
@@ -624,12 +624,6 @@ privilege mode and not in other modes. This is because interrupts for
 lower privilege modes are always disabled, whereas interrupts for higher
 privilege modes are always enabled.
 
-If the hart is currently running at some privilege mode `*x*`, an {ret}
-instruction that changes the privilege mode to a mode less
-privileged than `x` also sets {intthresh} to the lowest supported {intthresh} value.
-This helps software avoid a higher privilege mode from having a non-minimum threshold while a lower
-privilege mode is running.
-
 NOTE: The anticipated use of threshold is to provide critical sections
 within code running at one privilege mode, not to selectively mask
 interrupts before running lower-privilege code.  If desired,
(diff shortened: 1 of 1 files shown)


smatherd (commented):
I agree it is redundant.  The xpil/xintstatus.xil text is found in the Returns from Handler section.   there is also xret text in the xintthresh section, in the xinhv section, .  should xret text be in a single location (move xintthresh xret text to the returns from handler section), or should xret text be next to each csr field?    Maybe we should remove the returns from handler section and keep this section?

christian-herber-nxp:
> I agree it is redundant. The xpil/xintstatus.xil text is found in the Returns from Handler section. there is also xret text in the xintthresh section, in the xinhv section, . should xret text be in a single location (move xintthresh xret text to the returns from handler section), or should xret text be next to each csr field? Maybe we should remove the returns from handler section and keep this section?

Good catch, definitely needs further clean up.

return from handler section is specific to Smclic/Ssclic.
xinththresh is part of S*clic, and should thus be in that section, rather than with xintthresh description.

xinhv ist specific to the hardware vectoring, which is an optional extensions.
The hardware vectoring extension would further refine the xret behavior.
I will try to work out something on Monday.
I think it would be good if each extension quite explicitly outlines the trap and return behavior, and it is not "hidden" with CSR definitions. We can than put a reference in the CSR definition to the respective section.

christian-herber-nxp:
i think we have everything covered now.

Issue #470, opened 2025-04-01
Redundancy in description of interrupt enable


christian-herber-nxp:
> Within a privilege mode `*_x_*`, if the associated global
> interrupt-enable {ie} is clear, then no interrupts will be taken in
> that privilege mode, but a pending-enabled interrupt in a higher
> privilege mode will preempt current execution.  If {ie} is set, then
> pending-enabled interrupts at a higher interrupt level in the same
> privilege mode will preempt current execution and run the interrupt
> handler for the higher interrupt level.

> NOTE: Existing RISC-V interrupt behavior is retained, where incoming
> interrupts for a higher privilege mode can preempt an active interrupt
> handler running in a lower privilege mode, regardless of global
> interrupt enable in lower privilege mode.

I believe only one needs to be kept. The note seems to be stating the same and is more fitting, as this is already defined behavior in M/S.

smatherd:
This is the key difference between clic and clint:

> If {ie} is set, then
pending-enabled interrupts at a **higher interrupt level** in the same
privilege mode will preempt current execution

in clint, if {ie} is set, any pending-enabled interrupt will preempt current execution.  The note does not mention that.

maybe change the text to combine them?
>Within a privilege mode *_x_*, if the associated global
interrupt-enable {ie} is clear, then no interrupts will be taken in
that privilege mode, but a pending-enabled interrupt in a higher
privilege mode will preempt current execution regardless of global interrupt enables, like existing RISC-V interrupt behavior. If {ie} is set, then
pending-enabled interrupts at a higher interrupt level in the same
privilege mode will preempt current execution and run the interrupt
handler for the higher interrupt level.

christian-herber-nxp:
My suggestion:

The existing interrupt behavior is retained, with the following addition:
Interrupts are only considered pending, if their pending bit is set and their level is greater-equal the current effective interrupt level.

Note: The effective term is essential here to also capture the threshold behavior.

PR #484, opened 2025-04-05
clean up redundancy in description of interrupt preemption

smatherd:
For issue #470.  clean up and try to be more precise with interrupt preemption text.  updated the preemption table since e was not defined.

### DIFF
--- a/src/clic.adoc
+++ b/src/clic.adoc
@@ -482,21 +482,20 @@ can preempt lower-numbered interrupt levels.  Interrupt level 0
 corresponds to regular execution outside of an interrupt handler.
 Levels 1--255 correspond to interrupt handler levels. 
 
-Incoming interrupts with a higher interrupt level can preempt an
-active interrupt handler running at a lower interrupt level in the
-same privilege mode, provided interrupts are globally enabled in this
-privilege mode.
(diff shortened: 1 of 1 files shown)


src/clic.adoc
```diff
@@ -698,10 +697,11 @@ The global interrupt-enable bit of the handler's privilege mode is
 cleared, to prevent preemption by higher-level interrupts in the same
 privilege mode.
 
-The overall behavior is summarized in the following table: the Current
+The overall behavior of interrupts with `clicintip[__i__]` and `clicintie[__i__]` asserted 
+is summarized in the following table: the Current
 `p/ie/il` fields represent the current privilege mode `P` (not
 software visible), interrupt enable `ie` =
-({status}.{ie} & `clicintie[__i__]`)  and interrupt
+{status}.{ie} and interrupt
```
christian-herber-nxp, line 704: it is just a hinch, but maybe we are missing something now with delegation and virtual interrupts in AIA.
One reason why I am in favor of specifying differential rather than overall behavior.

christian-herber-nxp: seems fine. i checked this table from AIA:
image
and this is not in conflict.

PR #501, opened 2025-04-14
update to ssclic clicintlvl

### LINKED ISSUE #486: clicintlvl
clicintlvl is defined in two places in the document hierarchy.
Its register definition needs to move into the section on indirect CSRs.

### LINKED ISSUE #496: Ssclic is missing level configuration
I am not sure if this is intentional or not. I would think S mode should be allowed to configure its own levels

smatherd:
closes #496, closes #486

adds s-mode access to clicintlvl using siselect and sireg, added to ssclic section.


christian-herber-nxp (approved):
lgtm

PR #502, opened 2025-04-14
493 repeat text on exception behavior

### LINKED ISSUE #493: Repeat text on exception behavior
> If there is an access exception on the table fetch, {epc} is written with the faulting address.  {tval} is either set to zero or written with the faulting address.

This line can be removed. it is just reiterating existing behavior.

christian-herber-nxp:


christian-herber-nxp:
When cleaning up the text, I noticed that some parts of the text required xinhv to be cleared by HW, and others did not.
Need to clarify what we need.

src/clic.adoc
```diff
@@ -1022,12 +1019,30 @@ function pointers.
 NOTE: The hardware vectoring bit {inhv} is provided to allow resumable
 traps on fetches to the vector table.
 
-When a trap is taken, the {inhv} bit is set by hardware to indicate if {epc} is the address of a table entry
-or cleared by hardware to indicate if {epc} is the address of an instruction.
```
christian-herber-nxp, line 1026: get this text back into the spec, also clarify at which mode the fields are written/read

christian-herber-nxp:
Zcmt: It is recommended that the second fetch be ignored for hardware triggers and breakpoints.

christian-herber-nxp:
inhv removed

PR #510, opened 2025-04-15
Removing RNMI and NMI sections in favor of a short note

### LINKED ISSUE #473: section on NMI and RNMI do not provide added value
No special interaction is defined.
The CLIC spec will be integrated with the privileged spec together with NMI and RNMI, no need to have this here.

christian-herber-nxp:


christian-herber-nxp:
@smatherd

PR #512, opened 2025-04-15
504 exceptions should be vectored with hw vectoring

christian-herber-nxp:
@smatherd


christian-herber-nxp:
move sync exception vectoring into separate extension

Issue #519, opened 2025-05-05
Dependence on AIA


christian-herber-nxp:
Do we depend on AIA? Or are we just compatible with CLINT and AIA?

christian-herber-nxp:
clarified: We depend on AIA (Smaia and Ssaia). In their simplest form, this is very similar to the basic scheme of Machine mode, with a number of WARL registers implemented as read only zero.

Issue #500, opened 2025-04-09
Why does each external interrupt source i have both a cliciprio[i] and clicintlvl[i]?


Silabs-ArjanB:
Section 2.6.2. cliciprio[i] defines a cliciprio for each interrupt. Section 3.1. CLIC Interrupt Input Control (clicintlvl) defines a clicintlvl[i] per interrupt. 

In contrast the AIA specification does only have iprio[i] for the non-external interrupts. The iprio value for the machine-level external interrupt must be read-only zero. 

What is the reason for the possibility to annotate both level and prio for external interrupts (as this was not possible for AIA)?

And an even more fundamental question: Why bother about cliciprio at all? Instead all interrupt sources could be hooked up to an interrupt controller that just performs prioritization based on clicintlvl so that we would not have to deal with the two step prioritization caused by the legacy CLINT architecture.

christian-herber-nxp:
I am not necessarily happy with how the level and prio is currently speced. 
Arm has an equivalent split between prio and level called group priority and subpriority.

One reason why having this split can be useful is to limit the worst case interrupt stack size (which is the sum of the largest possible stacks per level). More levels means bigger stack.
Do you agree this makes sense?

Silabs-ArjanB:
In a previous version of the CLIC spec (version v0.9, 2024-06-28: Draft) there was the concept of 'interrupt priority' and 'interrupt level'. In that spec (only) level affects preemption and priority could be used to prioritize between interrupts at the same level. In that CLIC spec priority and level were **both** used to pick the most important interrupt to present to the core. I was completely fine with that proposal. That proposal was not related to AIA iprio in any way.

The current CLIC spec seems to have changed that in a way that only the level is used to pick the most important interrupt to present to the core. cliciprio gets factored in later, which is weird as then it might be too late (at only 1 external interrupt will be presented to the core and that might not have the highest cliciprio for a given level).

I agree it makes sense to limit the number of levels (levels of preemption), but I am really puzzled at why cliciprio was introduced that seems to match AIA iprio for some interrupts. If you want something like this then you need three concepts I think: level to be used to rank external interrupts, prio to rank external interrupts, prio for non external interrupts to rank against the one external interrupt presented to the core. Again, why use AIA and AIA iprio at all, it just seems to complicate things.

christian-herber-nxp:
TG Meeting:
- Specify if a partial implementation of lvl implements LSBs or MSBs.

[...]

PR #524, opened 2025-05-06
add a separate AIA compatibility extension

smatherd:
AIA adds additional complexity that some embedded systems may not need.  This extension moves those requirements to a separate extension.  For issues #519, #500

### DIFF
--- a/src/clic.adoc
+++ b/src/clic.adoc
@@ -8,7 +8,7 @@
 
 
 [[riscv-doc-template]]
-= AIA Core-Local Interrupt Controller (CLIC) RISC-V Privileged Architecture Extensions
+= Core-Local Interrupt Controller (CLIC) RISC-V Privileged Architecture Extensions
 include::../docs-resources/global-config.adoc[]
 :docgroup: RISC-V Task Group
 :description: RISC-V Example Specification Document (Zexmpl)
@@ -128,7 +128,7 @@ Graphics used are either explicitly available for free, are property of RISC-V I
 [Preface]
 == Copyright and license information
 
-This RISC-V AIA CLIC extension specification is © 2018-2025 RISC-V international
+This RISC-V CLIC extension specification is © 2018-2025 RISC-V international
 
 This document is released under a Creative Commons Attribution 4.0
 International License. +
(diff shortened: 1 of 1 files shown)


src/clic.adoc
```diff
@@ -160,12 +160,34 @@ This table provides a summary of the CLIC extensions to AIA.
...
 
 NOTE: The extensions defined here are orthogonal to the NMI and RNMI
 machanisms. Their behavior is unchanged by the extensions of CLIC.
 
-== Increase of AIA local interrupts - smclicincr
+== Compatibility with AIA - Smclicaia
+
+AIA has more complicated default priorities and additional restrictions.  
+
+=== CLIC Interrupt Attribute (`cliciprio`)
```
christian-herber-nxp, line 173: why does every section start with CLIC Interrupt Attribute?

christian-herber-nxp: actually I see it is just the first two.

src/clic.adoc
```diff
@@ -160,12 +160,34 @@ This table provides a summary of the CLIC extensions to AIA.
...
+Each interrupt has an associated priority as defined in the AIA specification.
+For the first 64 interrupts, these registers mirror the values of iprio registers in the AIA specification.
+
+=== CLIC Interrupt Attribute (`clicmideleg`)
+Each interrupt has an associated interrupt privilege mode delegation as defined in the AIA specification.
+For the first 64 interrupts, these registers mirror the values of mideleg bits in the AIA specification.
+
+=== CLIC Interrupt Enable (`clicintie`)
+When XLEN = 64, for the first 64 interrupts, these bits mirror the values of mie CSR.
+When XLEN = 32, for the first 32 interrupts, these bits mirror the values of mie CSR.
```
christian-herber-nxp, line 183: doesnt aia have 64 ie bits for XLEN 32?

src/clic.adoc
```diff
@@ -160,12 +160,34 @@ This table provides a summary of the CLIC extensions to AIA.
...
 | Sscspsw      | Conditional stack pointer swap at supervisor level
+| Smclicaia    | Compatibility with AIA
 |===
 
 NOTE: The extensions defined here are orthogonal to the NMI and RNMI
 machanisms. Their behavior is unchanged by the extensions of CLIC.
 
-== Increase of AIA local interrupts - smclicincr
+== Compatibility with AIA - Smclicaia
+
```
christian-herber-nxp, line 170: The Smclicaia extension depends upon the Smaia extension

christian-herber-nxp (changes requested):
does not lgtm right now.
The Smclicaia cannot be understood without reading the subsequent extension.
Also, it is not clear to me how this would interact with CLINT. Is the intention to mirror also? Doesn't seem to be in the spec.

What is the complexity that we do not want?
mie/mip/mideleg are not really changed with AIA, they are just widened for XLEN=64.
The prios only exist for AIA, so mirroring is only there for AIA.

This seems to be the only differences, and this should be fairly transparent and does not require a separate extension.

smatherd:
I'd like to close this pull request and replace it with #528.  I think it makes more sense to keep the other CLIC extensions compatible with AIA but add an option for implementations to assign CLIC interrupts generically that do not need AIA or legacy RISC-V mie/mip behavior.

PR #528, opened 2025-05-12
add a generic CLIC mode not compatible with AIA

smatherd:
For issues #519, #500.  AIA adds additional complexity that some embedded systems may not need.  This creates a CLIC mode, similar to previous CLIC specs that is incompatible with AIA/RISC-V priv interrupt scheme but allows implementations to generically assign interrupts.

### DIFF
--- a/src/clic.adoc
+++ b/src/clic.adoc
@@ -8,7 +8,7 @@
 
 
 [[riscv-doc-template]]
-= AIA Core-Local Interrupt Controller (CLIC) RISC-V Privileged Architecture Extensions
+= Core-Local Interrupt Controller (CLIC) RISC-V Privileged Architecture Extensions
 include::../docs-resources/global-config.adoc[]
 :docgroup: RISC-V Task Group
 :description: RISC-V Example Specification Document (Zexmpl)
@@ -127,7 +127,7 @@ Graphics used are either explicitly available for free, are property of RISC-V I
 [Preface]
 == Copyright and license information
 
-This RISC-V AIA CLIC extension specification is © 2018-2025 RISC-V international
+This RISC-V CLIC extension specification is © 2018-2025 RISC-V international
 
 This document is released under a Creative Commons Attribution 4.0
 International License. +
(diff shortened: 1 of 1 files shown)


christian-herber-nxp:
@dansmathers : what exactly do you view as too costly in Smaia for small systems? I polled this at the F2F today and put of 20-30 people, nobody seemed concerned with making Smaia mandatory as the base of CLIC.

smatherd:
In my implementation, calculating the priority encoding tree takes quite a few levels of logic.  AIA adds this default tie-breaking in table 8.  highest 43,11,3,7,9,1,5,12,10,2,6,13,35 lowest.  in the future there may be 23, 45, 17 as even lower priorities.

it also reserves the allocation of the interrupts so that only 24-31 and 48 and above are designated for custom use.   

The risc-v priv spec puts additional restrictions on the behavior of SEIP.  On only that interrupt, SEIP is the logical-OR of the software-writable bit and an interrupt signal from an external interrupt controller.  SSIP, LCOFIP are also unique in behavior.

In AIA, iprio2 bits 15:8 have a special requirement when supervisor external interrupts trap to M-mode.  

In AIA, If bit 9 for a supervisor external interrupt (SEI) is one in mideleg or mvien and in mvip, causing
sip.SEIP to be one, but there is no supervisor-level interrupt from the hart’s external interrupt
controller (APLIC or IMSIC), then a priority number for the SEI is not supplied by the external
interrupt controller as usual. In that case, the SEI is assigned a priority number of 256.

It's a lot of legacy and extra logic for implementation that may not provide any benefits, especially on a potentially critical timing path.

smatherd:
I see a table 8 in section 5.1 in Version 1.0, Revised 20250312
"Table 8. The standard major interrupt codes, listed in default priority order"

The big difference from the previous pull #524 is that the assumption is except for the smcliconly extensions, CLIC _is_ compatible with AIA.  These new extensions just provide an off-ramp for implementations that don't need the extra default AIA priority complexity and other mip/mie CSR quirks.  If smcliconly isn't supported, everything else is expected to be fully compatible with Smaia and that allows running CLINT/CLIC modes and hypervisor support.   

smcliconly is more restrictive in that all priv modes with interrupts need to be in clic mode, but then it's pretty much the straightforward clic generic interrupt selection tree as before in previous CLIC specs.  So for simpler designs or designs that want to assign interrupts without reserving space for things they'll never use.  

Let's discuss in the next TG so I can understand your concerns better.

christian-herber-nxp:
I see. Seems the AIA spec was updated without changing the version number, which changed all the table indeces.

Machine mode also has the priorities defined, so I am still not sure why AIA adds complexity in this regard.

Issue #489, opened 2025-04-07
clicintlvl: need to specify value for unimplemented bits


mark-honman:
The spec currently states:
A parameterized number of upper bits in clicintlvl[i] are assigned to encode the interrupt level.

However it does not specify what value the lower bits should read as. 
This was previously specified as "read as all-ones" which, while it meant efficient use of the level-encoding space, was hard for firmware users to understand. 

I think this boils down to choosing whether the lest significant bit should read as 0 or 1 when not implemented. The read-as values of the other unimplemented bits could be left unspecified.

PR #529, opened 2025-05-12
recommend unimplemented LSBs of clicintlvl be 1.

smatherd:
For issue #489, leave requirements of the number of levels to future platforms and profiles but add recommendation of LSBs of clicintlvl be 1 to be compatible with previous CLIC recommendations and to possibly make HW compatible with future platforms/profiles.


smatherd:
closing due to #533

PR #532, opened 2025-05-14
522 add mechanism analogous to claimi for sw vectored interrupts

christian-herber-nxp:
Adding the necessary registers to make the increase work.
I would also suggest we address the prio/level topic as part of this PR, but I will leave this a separate topic for now.

### DIFF
--- a/src/clic.adoc
+++ b/src/clic.adoc
@@ -82,6 +82,8 @@ Graphics used are either explicitly available for free, are property of RISC-V I
 :mpintstatus: pass:q[``mpintstatus``]
 :mintstatus: pass:q[``mintstatus``]
 :mintthresh: pass:q[``mintthresh``]
+:mtopi: pass:q[``mtopi``]
+:mclaimi: pass:q[``mclaimi``]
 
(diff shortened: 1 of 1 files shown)


src/clic.adoc
```diff
@@ -397,6 +403,37 @@ In this `miselect` offset range:
...
+----
+
+==== Claim top interrupt (mclaimi)
+
+Register {mclaimi} has the same value as {mtopi}.
+When this value is not zero,
+reading {mclaimi} has the simultaneous side effect of clearing the pending bit for the reported interrupt identity,
+if possible.
+Writes to {mclaimi} are ignored.
+
```
christian-herber-nxp, line 422: put a todo to check if side effect on read is acceptable, otherwise change to write access

christian-herber-nxp: There are no read-only accesses to CSRs anyway. The only limitation on side effects is that side effects should not occur if the dest is x0.
As writes are ignored according to the spec, you can use any flavor of CSRRRW/CSRRS/CSRRC or the immediate versions. All will have the same result.

PR #533, opened 2025-05-15
525 proposal remove levels to be more consistent with existing specs and terminology

christian-herber-nxp:


src/clic.adoc
```diff
@@ -898,36 +695,48 @@ spintstatus
...
+The behavior of an {sret} instruction is modified as follows:
+
+In CLIC mode, {sret} sets {sintstatus}.{sipprio} to {spintstatus}.{spipprio}.
+
 ==== State Enable
 
 If the Smstateen extension is implemented, then the bit 53 (CLIC) in mstateen0 is
 implemented. If bit 53 (CLIC) of a controlling mstateen0 CSR is zero, then
-access to the new CSRs ({sivt}, {sintthresh})
+access to the new CSRs ({sivt})
```
tovine, line 739:
```suggestion
access to the new CSR ({sivt})
```
Changed from plural to singular since there's only one CSR left in the list

src/clic.adoc
```diff
@@ -639,22 +571,34 @@ where:
...
+The interrupt preemption priority {ipprio} is defined as follows:
+
+[source]
+----
+IPPRIO_MASK = 2^(mipreemptcfg.preemptmsk) - 1
+ipprio[i]   = iprio[i] & IPPRIO_MASK
+----
+
+A nested horizontal preemption will happen if and only if an interrupt i is pending and enabled,
+and the preemption priority of that interrupt ipprio[i] excceds (i.e. is numerically smaller) than {mintstatus}.{mipprio}.
```
tovine, line 596: Just to make sure: this is essentially a renaming, and the existing functionality of mulitple interrupt levels and preemption is preserved?

src/clic.adoc
```diff
@@ -664,226 +608,79 @@ Bits   Field         Description
...
  bits    description
- 31:24   mil
+ 31:24   mipprio
  23:16   (reserved)
- 15: 8   sil if ssclic is supported
+ 15: 8   sipprio if ssclic is supported
   7: 0   (reserved)
 ----
 
-==== New Interrupt-Level Threshold ({mintthresh}) CSRs
```
tovine, line 691: Is this threshold functionality removed?
If so, how can software implement "semi-critical" sections where most interrupts are masked but high importance ones are let through?

src/clic.adoc
```diff
@@ -639,22 +571,34 @@ where:
...
+----
+
+{mipreemptcfg} is a WARL register used to configure the behavior of preemption.
+{mipreemptcfg}.{preemptmsk} configures the number of least significant bits of iprio masked out for constructing {ipprio}.
+The highest legal value of {mipreemptcfg}.{preemptmsk} is IPRIOLEN.
+The interrupt preemption priority {ipprio} is defined as follows:
+
+[source]
+----
+IPPRIO_MASK = 2^(mipreemptcfg.preemptmsk) - 1
```
christian-herber-nxp, line 591: TODO: review the logic

src/clic.adoc
christian-herber-nxp: yes. it is going back to how previous versions of the spec worked, but using preemption priority instead of level.
In contrast to level, for the preemption priority, a smaller value has priority over a small value, to make it consistent with AIA and any other RISC-V spec.

src/clic.adoc
christian-herber-nxp: I removed this here, because i added a priority threshold to the Smclicincr extension in another PR.
Such a mechanism is needed with a high number of sources, independent of whether nested preemption is supported or not.

src/clic.adoc
christian-herber-nxp: fixed

src/clic.adoc
christian-herber-nxp: fixed

christian-herber-nxp:
from my point of view, this looks good to go.

src/clic.adoc
```diff
@@ -525,79 +527,12 @@ When {sithreshold} is a nonzero value P,
 interrupt sources with priority numbers P and higher behave as though those sources were not enabled,
 regardless of the settings of their interrupt-enable bits.
 
-== Same Privilege Mode Interrupt Preemption Support - smclic
+== Horizontal Nested Interrupt Preemption support for M-mode - smclic
 
-The Smclic extension depends on the Smclicincr extension.
+The Smclic extension depends on the Smaia extension.
```
tovine, line 532: I guess this dependency is only to pull in these CSRs, and doesn't require implementing the rest of the AIA?
bilde

tovine (approved):

src/clic.adoc
christian-herber-nxp: Smaia and Ssaia are clearly defined in the AIA spec. This does not include things like the APLIC. It should also be noted that many CSRs above are WARL, so zero is a valid implementation of some of them, making it very similar to the original behavior in machine mode.

smatherd:
1. can we change the names ipprio  and pipprio to be something pronounceable and something that is more than a single typo away from iprio?   are you sure you want to reuse the letter p for priority and preemption and previous?  

2. is preemptmsk supposed to be 3:0 instead of 3:1 so it can be set from 0 to 8?  if not, do you need to define bit 0 as reserved?
> 3:1         preemptmsk

3. This pull removes a lot of the WFI behavior details.  can we provide some notes on how xithreshold affects higher privilege preemption and WFI?

> When mithreshold is a nonzero value P, interrupt sources with priority numbers P and higher behave as though those sources were not enabled, regardless of the settings of their interrupt-enable bits.

In previous versions of CLIC if the hart was running in s-mode and an m-mode interrupt was pending and enabled, mithreshold was ignored and the m-mode interrupt was taken.  with these spec revisions mithreshold now keeps the m-mode interrupt from being taken?   
In previous versions of CLIC, if the hart was running in m-mode and an s-mode interrupt was pending and enabled, WFI was required to wake up ignoring sithreshold.  does this pull imply that sithreshold would now be considered when determining if the hart is required to wake-up?

4. When setting preemption priority to IPPRIO_MAX, is that equivalent to behaving as though those sources were not enabled for both horizontal and vertical interrupts and for WFI?
> Interrupt preemption priority IPPRIO_MAX corresponds to regular execution outside of an interrupt handler

christian-herber-nxp:
> 1. can we change the names ipprio  and pipprio to be something pronounceable and something that is more than a single typo away from iprio?   are you sure you want to reuse the letter p for priority and preemption and previous?

I am very open to suggestions. It is just what came to mind.

> 2. is preemptmsk supposed to be 3:0 instead of 3:1 so it can be set from 0 to 8?  if not, do you need to define bit 0 as reserved?

can be set from 0 - 7. The corner cases are: prio = preemption prio, and all preemption prios are equal, i.e. no nested preemption.
I think is is ok to have these, and no special treatment is needed.

> > 3:1         preemptmsk
> 
> 3. This pull removes a lot of the WFI behavior details.  can we provide some notes on how xithreshold affects higher privilege preemption and WFI?

ok

> > When mithreshold is a nonzero value P, interrupt sources with priority numbers P and higher behave as though those sources were not enabled, regardless of the settings of their interrupt-enable bits.
> 
> In previous versions of CLIC if the hart was running in s-mode and an m-mode interrupt was pending and enabled, mithreshold was ignored and the m-mode interrupt was taken. with these spec revisions mithreshold now keeps the m-mode interrupt from being taken? In previous versions of CLIC, if the hart was running in m-mode and an s-mode interrupt was pending and enabled, WFI was required to wake up ignoring sithreshold. does this pull imply that sithreshold would now be considered when determining if the hart is required to wake-up?
> 
> 4. When setting preemption priority to IPPRIO_MAX, is that equivalent to behaving as though those sources were not enabled for both horizontal and vertical interrupts and for WFI?

I think it would be a strange setting. But I believe the text is currently in a way, that the preemption priority is only considered for nested preemption.

> > Interrupt preemption priority IPPRIO_MAX corresponds to regular execution outside of an interrupt handler

christian-herber-nxp:
> In previous versions of CLIC if the hart was running in s-mode and an m-mode interrupt was pending and enabled, mithreshold was ignored and the m-mode interrupt was taken. with these spec revisions mithreshold now keeps the m-mode interrupt from being taken?

what was the reason as to why that was? I am open to this, but i would like to understand what the benefit is.
I think that if you are returning to a less privileged mode, you should have put mithreshold to the minimum anyway, so I do not see much of an impact.

christian-herber-nxp:
> > 1. can we change the names ipprio  and pipprio to be something pronounceable and something that is more than a single typo away from iprio?   are you sure you want to reuse the letter p for priority and preemption and previous?
> 
> I am very open to suggestions. It is just what came to mind.

Made a change to use nested interrupt preemption priority: nipprio

How does that sound?

src/clic.adoc
```diff
@@ -639,295 +571,176 @@ where:
...
+{mipreemptcfg} is a WARL register used to configure the behavior of preemption.
+{mipreemptcfg}.{preemptmsk} configures the number of least significant bits of iprio masked out for constructing {nipprio}.
+The highest legal value of {mipreemptcfg}.{preemptmsk} is IPRIOLEN.
+The nested interrupt preemption priority {nipprio} is defined as follows:
+
+[source]
+----
+NIPPRIO_MASK = ~(2^(mipreemptcfg.preemptmsk) - 1)
+nipprio[i]   = iprio[i] & NIPPRIO_MASK
+----
```
smatherd, line 593: bit 0 of mipreemptcfg isn't currently defined.   

It says the highest legal value of preemptmsk is IPRIOLEN but in AIA spec, IPRIOLEN can be 8 but preemptmsk can only take the values 0 thru 7.

PR #535, opened 2025-05-23
Removing CLIC mode

### LINKED ISSUE #534: Change CLIC mode
Currently, a new mode in mtvec is introduced for CLIC mode.
CLIC mode allows switching dynamically between CLINT and CLIC mode, where CLIC mode
- changes the vectoring behavior
- activates the special trap and return behavior needed for preemption

I do not see value in switching dynamically.
To me, it would be fine to have the behavior coupled to having the extension implemented.
There is not a lot of dynamically selectable/switchable extensions in RISC-V.

My proposal would be as follows:
- Smclic also works with mtvec.mode={0,1}

We either drop mtvec.mode=2, or mtvec.mode=2 is a mode which is equivalent to mode=1 (vectored), but this mode uses handler addresses in the vector table instead of instructions (mtvt). This would also address https://github.com/riscv/riscv-fast-interrupt/issues/466.

christian-herber-nxp:
It is replaced by an interrupt vector table mode, which uses the ivt CSR All other modes are now accessible with all CLIC extensions, making backwards compatibility easy

### DIFF
--- a/src/clic.adoc
+++ b/src/clic.adoc
@@ -161,7 +161,7 @@ This table provides a summary of the CLIC extensions.
 | ssclicincr   | Increase of the number of local interrupts for S-mode
 | smclic       | Horizontal Nested Interrupt Preemption support for M-mode
 | ssclic       | Horizontal Nested Interrupt Preemption support for S-mode
-| smclichv     | Hardware Vectored Interrupts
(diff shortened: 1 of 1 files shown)


smatherd:
this looks fine to me.

PR #552, opened 2025-06-20
248 hypervisor mode should be discussed

christian-herber-nxp:


christian-herber-nxp:
we will continue the discussion. This does not seem to be the solution

PR #556, opened 2025-06-24
Removing side effect on CSR read for claimi

### LINKED ISSUE #555: Clarify side effect on read for xclaimi CSR

christian-herber-nxp:


allenjbaum (commented):
This change has a dependency on mtopi - yet mtopi is defined nowhere that I can find in this adoc

allenjbaum (commented):
This says "when this value is not zero" which leaves "this value" to be a bit undefined. 
"When the value of {mmtopi} is not zero, any write to {mclaimi} has only the simultaneous side effect of clearing the pending bit for the reported interrupt identity

Under what conditions would clearing the pending bit not be possible?

allenjbaum (changes requested):
Same comments as the above for mclaimi

christian-herber-nxp:
I am closing this PR, as I believe this part will anyway be reworked significantly. 
The correct way to do it would be as IMSIC does it, and have reads to mtopi clear the ip bit. It is split in APLIC into topi and claimi because this is memory mapped.

christian-herber-nxp:
> My concern was that CSR reads aren't supposed to have side effect on the same CSR. If AIA already has reads perform side effects on a different CSR, that's OK.

it is a CSR write that has the clearing effect.
On the memory mapped register, reading has the side effect.

allenjbaum:
Ah, I missed that one was MMIO. Thx.

PR #564, opened 2025-08-11
Address earls input

christian-herber-nxp:


christian-herber-nxp:
merging this as there was additional input on this paragraph. trying to avoid conflicts.

Issue #605, opened 2025-09-15
Major and external interrupts unification [proposal]


evgeniy-paltsev:
Current ACLIC spec has two interrupt types: major and external interrupts.

To achieve interrupts nesting and vectoring for both interrupt types the ACLIC spec introduces several new CSRs and extensions:
- new `mivt`, `sivt`, `meivt`, `seivt` CSRs for major and external IVTs
- new Smaiae extension to provide
  - `stopi`, `mtopi` CSRs which needed for information about priority and identity of current top major interrupt
  - `iprio0-15` iCSRs array to have configurable priority for major interrupts

We want to propose an alternative approach, which is that we unify major and external interrupts:
- In ACLIC mode all major interrupt sources are forwarded to ACLIC, according to their identity numbers. CLINT is not used at all and ACLIC becomes the only source of interrupts for M/S-mode.
- When ACLIC mode is disabled (legacy mode) major interrupts are routed to CLINT (as in RISC-V ISA). In that mode there is no configurable priority for major interrupts, no vectoring, no nesting.

The resource-constrained systems targeted by ACLIC are likely to have limited number of major local interrupt sources (M-timer, S-timer, M-software interrupts). Among all of them, only the M-timer interrupt is mandatory, according to the RISC-V ISA spec. So the resources duplication between CLINT and ACLIC is minimal.

Proposed approach has several advantages:
- we don't need multiple IVTs and `mivt`, `sivt`, `meivt`, `seivt` CSRs. Existing `xtvec` CSRs can be used for vector base.
- no need in `mtopi`, `stopi` CSRs
- no need in `iprio0-15` iCSRs array
  - we can avoid introducing of new `Smaiae` extension and drop AIA dependency
    - if we drop AIA dependency we can also drop `miph`, `siph`, `mieh`, `sieh`, `midelegh` CSRs
- we simplify interrupt trap logic
- we simplify interrupt configuration logic - all interrupts are configured the same way
- we simplify `push` logic of tpush extension
- still fully compatible with RISC-V ISA
- trivial further size reduction:
  - i.e.: application-specific IC with stricter resource constraints (and without requirement of being fully compatible to RISC-V) can implement ACLIC without CLINT at all. No ACLIC architectural changes needed for that, HW still be fully compatible with ACLIC handling mode

Since this approach offers a minimalistic ACLIC implementation, it may be more beneficial to use it as the default rather than as a separate extension.

[...]

PR #606, opened 2025-09-15
Major and external interrupts unification

evgeniy-paltsev:
This PR unifies major and external interrupts by forwarding all major interrupts to ACLIC when the ACLIC mode is selected. The additional technical details and discussion can be found in the #605 issue.

Fixes: #605

### DIFF
--- a/src/aclic.adoc
+++ b/src/aclic.adoc
@@ -1099,18 +1099,10 @@ push:
   sr a4, OFFSET(sp)         # Save xepc
   sr a5, OFFSET(sp)         # Save xpistatus
   bgez a0, handle_exc       # Handle synchronous exception
-  csrr a0, xtopi            # Read major interrupt number into a0
-  li a1, xEI                # Write major identity of external interrupt to a1
-  beq a0, a1, handle_ei     # Handle external interrupt
-  # clear major interrupt pending bit - no single instruction equivalent
-shift_idenity:
+  csrrw a0, xtopei, zero    # Read interrupt number into a0 and clear pending bit
   srli a0, a0, 16           # Right shift topi value to have idenity aligned with bit 0.
 finish_push:
   csrrsi zero, xstatus, XIE # Enable interrupts
-
-handle_ei:
-  csrr a0, xtopei           # Read minor interrupt number into a0
-  j shift_identity           # Jump back to rest of push sequence
 
 handle_exc:
   csrr a1, xtval            # Get xcause of interrupted context


mark-honman:
On the face of it this PR appears to remove vectoring of the minor interrupts, while retaining the minor interrupt concept which is foundational to AIA. This does not make sense.

src/aclic.adoc
```diff
@@ -413,6 +407,25 @@ Therefore, the registers setipnum, clripnum, setienum, clrienum, setipnum_be, an
 The iprio portion of it can be accessed as described in the section on indirect access.
 The rest is not required.
 
+=== Major interrupt redirection to ACLIC
+
+In ACLIC mode, ACLIC becomes the hart's primary - and only - interrupt controller. The major interrupt sources (i.e, timer interrupt, software interrupt, ...) are forwarded to ACLIC, according to their major identity numbers.
+
+NOTE: all major interrupts keep their identity numbers.
+
+In this mode CLINT is not used and doesn't influence the interrupt handling.
```
christian-herber-nxp, line 416: CLINT is not really a thing in the world of RISC-V standards

christian-herber-nxp:
high level comment: I can see small benefits on HW side, while I do also concern about SW (suddenly, and OS kernel will rely on a driver for external interrupt controller to be able to e.g. set the SW interrupt).
This should be traded off against simply restricting the vectoring to the external interrupts - in my considerations, HW-vectoring is not important for major identities.

jackwdandrew:
This PR was presented by Evgenii to Fast Interrupts TG on 29th September 2025.

@evgeniy-paltsev, could you attach the slides?

christian-herber-nxp:
was there any input from the TG?
Is there support other than Synopsys for this?

christian-herber-nxp:
what is the layout of the unified vector table? First 64 entries reserved for major interrupts? Platform defined?

evgeniy-paltsev:
> This PR was presented by Evgenii to Fast Interrupts TG on 29th September 2025.
> @evgeniy-paltsev, could you attach the slides?

In addition to sharing the slides via the tech-fast-int mailing list, I'm attaching them here as well.
[Interrupts unification proposal.pdf](https://github.com/user-attachments/files/22743813/Interrupts.unification.proposal.pdf)

evgeniy-paltsev:
Hi @christian-herber-nxp,

> This should be traded off against simply restricting the vectoring to the external interrupts - in my considerations, HW-vectoring is not important for major identities.

Could you clarify this a bit? Am I correct in understanding that you're suggesting a single entry point for all major interrupts?

> I do also concern about SW (suddenly, and OS kernel will rely on a driver for external interrupt controller to be able to e.g. set the SW interrupt).

I'm not sure I fully follow your concern here. If the new HW features aren't used, then no SW changes are needed. But if they are used (e.g., interrupt nesting), then naturally the SW must be updated to support them.

christian-herber-nxp:
> > This should be traded off against simply restricting the vectoring to the external interrupts - in my considerations, HW-vectoring is not important for major identities.
> 
> Could you clarify this a bit? Am I correct in understanding that you're suggesting a single entry point for all major interrupts?

Yes. This was previously floated by Mark and makes a ton of sense.

> > I do also concern about SW (suddenly, and OS kernel will rely on a driver for external interrupt controller to be able to e.g. set the SW interrupt).
> 
> I'm not sure I fully follow your concern here. If the new HW features aren't used, then no SW changes are needed. But if they are used (e.g., interrupt nesting), then naturally the SW must be updated to support them.

Everything CAN be supported in SW - so is the nature of software. But for maintenance, quality, ramp-up purposes it is preferable to have architecture support that is not full of ifdefs for different flavors of the architecture. A new feature is a new feature and needs additional code to be supported.
Making major interrupts external is not really a feature by itself, it is just changing existing behavior to have a slightly more efficient HW implementation. Still, it will need additional ifdef-else macros in the base port of risc-v and the ability to redirect e.g. writes to the pending bit of a SW interrupt to either the traditional bit, or a bit managed by the external interrupt controller, which will be modelled as a device driver rather than be considered part of the core.
It can be done, of course, but the question is if the benefit is worth it.

evgeniy-paltsev:
> what is the layout of the unified vector table? First 64 entries reserved for major interrupts? Platform defined?

Entry 0 - pointer to exception handler
Entries 1...N - pointers to interrupt handlers, where each index corresponds to the interrupt identity number

The major interrupts retain their original identity numbers, so their positions in the vector table are fixed. The remaining identity numbers (those not assigned to implemented major interrupts) are available for use by external interrupts.
There's no requirement for platforms to reserve the first 64 entries (or any specific range) for major interrupts, though they are free to do so if desired.

PR #609, opened 2025-09-18
Implement ACLIC as a subset of APLIC interface

### LINKED ISSUE #572: Unnecessary dependency on Smaia/Ssaia
As per RISC-V AIA, the APLIC domain is the interrupt configuration interface for the group of physical cores running the same privilege level. 
The core's privilege may or may not have permissions to access APLIC domain configuration and/or receive the interrupts delivered by messages or directly (mutually exclusive at domain level). 

APLIC direct interrupt delivery is intended to supply the cores that are not aware of AIA. In such case there is no AIA specific support is required at the core side except the ability to receive single wired interrupt. This means that there is no dependency on Smaia and Ssaia for APLIC/IDC based configurations.  

To control the interrupts APLIC provides the IDC (interrupt delivery control) MMIO interface individual for the privilege level (M or S, but not VS) for each physical core.  In case of single core, the APLIC may have two domains (M and S) and provide IDC interfaces for M-mode and S-mode privileges (direct delivery to VS-mode is not assumed by AIA). 

Mixing of APLIC domain configuration, APLIC IDC, and IMSIC interfaces (as it is documented now) looks incorrect and confusing. Instead it would be reasonable to reuse only APLIC domain configuration and IDC structure as is, improving by migration to iCSRs instead of MMIO. IMSIC interface should not be inherited/exposed in such case and dependency on Ssaia/Smaia looks as unnecessary overhead for single core.

evgeniy-paltsev:
This PR proposes ACLIC interface based on subset of AIA Advanced Platform Level Interrupt Controller (APLIC).

In pair with #606 this resolve #572

### DIFF
--- a/src/aclic.adoc
+++ b/src/aclic.adoc
@@ -209,8 +209,8 @@ Each row contains the dependencies of the extension named in the first column.
 | Sscsps         |    | x  |          |          |        |  
 | Smtp           | (x)| (x)|          |          |        |  
 | Smaiae         | x  | (x)| x        | (x)      |        |  
-| Smaclic        |    |    |          |          | x      |  
-| Ssaclic        |    |    |          |          | x      |  
+| Smaclic        |    |    |          |          |        |
+| Ssaclic        |    |    |          |          |        |
 | Smnip          |    |    |          |          | x      |  
 | Ssnip          |    |    |          |          |        | x
 | Smehv          |    |    |          |          |        | x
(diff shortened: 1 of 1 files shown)


mark-honman:
I'm confused... the change in the PR removes the Smivt/Ssivt extensions but does not make any changes to the ACLIC interface?

mark-honman:
You got it! I had cliced on the #606 link in the head comment, sorry!

PR #614, opened 2025-09-22
612 add fs and vs bits to xpistatus

### LINKED ISSUE #612: Add FS and VS bits to xpistatus
mstatus.{FS, VS} should be mirrored into xpistatus to assist context saving.
Additionally, the push instruction should set both VS and FS to 0 after saving the context, to ensure an ISR using the FPU or vector unit without saving the context prior (and setting these flags to clean) will cause an illegal instruction exception.
No modification needed for pop instruction.

christian-herber-nxp:


evgeniy-paltsev:
> SD is a weak argument. It is a read only field. 

`SD` is read-only, but unfortunately it doesn't mean that we can just ignore it.

When the `Smtp` extension present, the `FS` and `VS` fields in `xstatus` will be restored by `tpopxret` instruction (which is the last instruction in the handler). However, we need the SD bit value (which depends on these bits) before the `tpopxret` if we've decided to do a context switch.

christian-herber-nxp:
> > SD is a weak argument. It is a read only field.
> 
> `SD` is read-only, but unfortunately it doesn't mean that we can just ignore it.
> 
> When the `Smtp` extension present, the `FS` and `VS` fields in `xstatus` will be restored by `tpopxret` instruction (which is the last instruction in the handler). However, we need the SD bit value (which depends on these bits) before the `tpopxret` if we've decided to do a context switch.

Restoring it can always be done, without a tpopxret instruction.
Of course, we could also consider adding a tpop instruction (leaving out the xret portion if that helps.

Issue #622, opened 2025-09-23
Remove Smaiae extension


mark-honman:
A careful reading of the AIA specification shows that mvip and mvien must not be implemented when the hart only supports M-mode.
If the hart supports S-mode, the minimum required implementation of mvip and mvien is not area intensive so there is no need for a dedicated extension to de-feature Smaia.

reosdev:
I agree that defining alternative AIA is the obvious overhead. However, I don't see the benefit in dependency on AIA. What AIA feature is beneficial for small cores? For now there is only reason to inherit priorities for few major interrupts. This is solved in alternative way with major interrupts redirection via external sources (only for nested vectored modes when configurable priorities are required).

mark-honman:
The point of this issue is that a dependency on Smaia does not add overhead in the case of mvip and mvien; that's what allows us to remove Smaiae which is currently defined to remove these registers. 

Here is what the AIA spec says:
When supervisor mode is implemented, the minimal required implementation of mvien and mvip has all bits being read-only zeros except for mvip bits 1 and 9, and sometimes bit 5, each of which is an alias of an existing writable bit in mip. (Although, as noted, it is strongly recommended that bit 9 of mvien also be writable.) When supervisor mode is not implemented, registers mvien and mvip do not exist.

reosdev:
The point of my comment is that dependency on AIA (or any variation of it) is not required at all, which includes registers you mentioned. Resource constrained interrupt controller should not depend on any extension that add complexity without benefits (read-only registers is good example). This does not mean ACLIC cannot coexist with AIA if this is required by some reason.

mark-honman:
Read-only zero registers really do not add significant complexity. The benefit is compatibility and re-use, both conceptual alignment with AIA and re-use of SW that has been written to interrogate the AIA mvip and mvien registers.

mark-honman:
Sure; it enables the same SW to be used for minimal and "full" implementations of mvip/mvien functionality. Since the minimal feature-set is defined in AIA itself, all SW that interacts with AIA HW must be prepared to deal with a minimal HW implementation. 
Such SW is fully re-usable with Smaclic HW that has a minimal implementation of mvip/mvien.

christian-herber-nxp:
> Could you please explain little bit more about benefits for SW in having read-only registers. In my understanding this is right way to confuse SW and add complexity in it. 
> 
> 

I think you are overplaying the complexity of a handful of read-only registers.
[...]

Issue #619, opened 2025-09-22
Drop "APLIC configured for ACLIC operation" section


mark-honman:
This seems to imply that it is possible for the system to contain an APLIC as well as one or more ACLICs. 
However the mechanism for wired-interrupt delivery from APLIC to ACLIC is not specified anywhere. 

Proposal: 
add MSI-receiving capability as an ACLIC extension (see issue #616 )

then unmodified APLIC can use MSI delivery to ACLIC in systems that require both types of interrupt controller to be present.

mark-honman:
I believe PR #662 addresses this source of the misconceptions, so closing the issue.

Issue #618, opened 2025-09-22
Indirect access to source and priority configuration


mark-honman:
with reference to section 3.2 "Indirect access to APLIC registers".

Suggest renaming to make clear that there is no APLIC. 

Rather say something like:

-  The ACLIC provides interrupt source and priority configuration capabilities as defined for the APLIC, except that when the source configuration Delegate bit is set, bits 9:0 identify the ACLIC interrupt file to which the interrupt is delegated. [or perhaps it is just a mode-delegation...]
- As with AIA hart-local interrupt pending and enable registers, the interrupt source and priority configuration are provided with an indirect CSR interface.

Issue #617, opened 2025-09-22
simplified eidelivery and IMSIC-like CSRs


mark-honman:
with reference to section 3.1 of the spec

- there is no provision for ACLIC and IMSIC to coexist, so the spec does not need to talk about IMSIC. Except possibly to say - "ACLIC provides an indirect eidelivery CSR, as in an AIA IMSIC".  
- since the ACLIC owns the eidelivery register, there is no need for an "interrupt delivery from ACLIC" eidelivery value (it would only be needed if the hart could have both an ACLIC and IMSIC - a possibility not covered by the spec). 
- let eip and eie be defined as they are for AIA IMSIC, call this "the AIA interface for hart-local external interrupt controllers"
- All the text explaining that eithreshold is an alias of ithreshold, eip relationship to setip and clrip etc. falls away because there are no APLIC like setip, clrip etc. registers (there is no description of those registers elsewhere in aclic.adoc)
-  Remove references to "connected APLIC domain" because there is no APLIC.

reosdev:
> there is no provision for ACLIC and IMSIC to coexist, so the spec does not need to talk about IMSIC. 

Totally agree. Current APLIC-based design should not talk about IMSIC at all.  

> since the ACLIC owns the eidelivery register, there is no need for an "interrupt delivery from ACLIC" eidelivery value (it would only be needed if the hart could have both an ACLIC and IMSIC - a possibility not covered by the spec).

This eidelivery redefinition is confusing. In case of IMSIC, delivery mode intended to manage connection of external interrupt controller (i.e. IMSIC, PLIC or APLIC in direct mode). This assumes that major external interrupt input can be switched between various external controllers. If ACLIC coexists with AIA IMSIC, then it makes sense to define such alternative for it. Otherwise, eidelivery is not required. 

> let eip and eie be defined as they are for AIA IMSIC, call this "the AIA interface for hart-local external interrupt controllers"

ACLIC pending and enable array definition is most confusing thing. In case of IMISC these arrays are indexed by priority (EIID) but not by interrupt source ID (IID). How ACLIC eliminate this difference? If not, then this is not IMSIC p&e bit arrays.

> All the text explaining that eithreshold is an alias of ithreshold, eip relationship to setip and clrip etc. falls away because there are no APLIC like setip, clrip etc. registers (there is no description of those registers elsewhere in aclic.adoc)

[...]

PR #625, opened 2025-09-23
Clarified AIA alignment

mark-honman:
Each of the 6 commits in the PR corresponds to one of the issues that I've recently opened to make the ACLIC spec more readable. 

#615 ACLIC is introduced as a 3rd AIA external interrupt controller option, with a hart local interface like an IMSIC but implementing APLIC like source configuration and interrupt prioritization. 

#617 An IMSIC like eidelivery register can be used without changes -  since eidelivery = 1 means delivery from the interrupt files inside a hart-local interrupt controller. Once the ACLIC is presented as an aliternative to IMSIC, rather than a heavily modified APLIC, the lengthly description of mapping between APLIC and IMSIC type registers is no longer required. 

#618 Indirect source and priority configuration is described as a way to merge the wired-interrupt configuration capabilities of APLIC, into a hart-local interrupt controller as an extension to the set of ei-prefixed indirect CSRs in an interrupt file. Register naming and addressing are changed to align with existing ei-regs BUT I have not checked that the suggested part of CSRind space is in fact available. Since ACLIC contains interrupt files rather than APLIC like domains, the more limited sourcecfg use model is explained in detail. 

#618 State enables - attempted to update this to match the renaming in the previous commit; however I'm not familiar with state enables and have probably got it wrong. 

#619 "APLIC configured for ACLIC operation" section is deleted, because it seems to be based on an early iteration of the spec in which ACLIC was more like a specialized APLIC. I found this section especially confusing because it was not clear whether the referenced APLIC was an actual APLIC, or a set of APLIC like registers inside ACLIC that are not described elsewhere. 

#622 Smaiae is removed as an extension in its own right; we found that AIA provides guidance on a minimal implementation of S-mode mvip and mvien that does not have an excessive area penalty. The AIA recommendations are noted in the Smaclic/Ssaclic extension description.


src/aclic.adoc
```diff
@@ -167,20 +167,30 @@ https://creativecommons.org/licenses/by/4.0/.
...
+
+This specification introduces a third type of AIA compliant interrupt controller. 
+
+The Advanced Core Local Interrupt Controller (ACLIC) is a single hart external interrupt controller optimized for real-time systems:
 
   * Low latency interrupt handling (achieved through nested preemption)
   * High throughput interrupt handling
   * Up to 1023 external interrupts
   * Low area overhead
-  * Maximize compatibility with AIA
```
christian-herber-nxp, line 176: why was this removed? Note that this still exists as bullet for Smaclic, as subgoals of the main goals

src/aclic.adoc
```diff
@@ -252,72 +242,62 @@ Of the main ACLIC goals, these targets are addressed here:
   * Maximize compatibility with AIA
 
 To achieve this, ACLIC builds upon the hart-level extensions of AIA,
-and the Advanced Platform Level Interrupt Controller (APLIC).
-The ACLIC is a tailored APLIC, optimized for resource efficiency, directly integrated with one hart.
```
christian-herber-nxp, line 256: it seems APLIC is mentioned twice?

src/aclic.adoc
```diff
@@ -252,72 +242,62 @@ Of the main ACLIC goals, these targets are addressed here:
...
+Since the ACLIC is closely coupled to a single hart for which AIA already defines an
+indirect CSR access mechanism, interrupt source and priority configuration is performed 
+using new indirect CSRs instead of the memory-mapped registers of an external APLIC. 
+The source and priority configuration data follow the same definitions as the equivalent APLIC fields. 
 
-When interrupts are delivered from an ACLIC, the following behavior is valid:
+Note: 
+  * A hart cannot have more than one closely coupled external interrupt controller; 
+  * in particular a hart cannot have both an ACLIC and an IMSIC. 
+  * A system can contain an APLIC as well as a closely coupled ACLIC; in this case {eidelivery} is
```
christian-herber-nxp, line 259: how does this selection work? You removed the eidelivery changes.

src/aclic.adoc
```diff
@@ -252,72 +242,62 @@ Of the main ACLIC goals, these targets are addressed here:
...
 
-The {eithreshold} CSR is an alias of the {ithreshold}.
-For the interrupt enable and interrupt pending bits {eipk} and {eiek},
-a write with the value 1 performs a write access with the value 1 to the corresponding {setip} or {setie} register,
-while a write with the value 0 performs a write access with value 1 to the corresponding {in_clrip} or {clrie} register.
-On read of {eipk} or {eiek}, these CSRs are an alias of {setip}[_k_] and {setie}[_k_], respectively.
+The number of interrupt identities supported by an ACLIC interrupt file is one less that a multiple of 64,
+unless the number of identities is less than 64. 
+Unlike the IMSIC for which the minimum number of supported identities is 63,
+an ACLIC can additionally support 7, 15, 31, or 63 interrupt identities.
```
christian-herber-nxp, line 271: Is this necessary? They could just be read-only zero. I.e. stick to the minimum of 64

src/aclic.adoc
```diff
@@ -252,72 +242,62 @@ Of the main ACLIC goals, these targets are addressed here:
...
-while a write with the value 0 performs a write access with value 1 to the corresponding {in_clrip} or {clrie} register.
-On read of {eipk} or {eiek}, these CSRs are an alias of {setip}[_k_] and {setie}[_k_], respectively.
+The number of interrupt identities supported by an ACLIC interrupt file is one less that a multiple of 64,
+unless the number of identities is less than 64. 
+Unlike the IMSIC for which the minimum number of supported identities is 63,
+an ACLIC can additionally support 7, 15, 31, or 63 interrupt identities. 
+In this case the {eip} and {eie} bits and source and priority configuration for the unimplemented
+identities are WARL 0. 
 
-For XLEN=32, {eipk} and {eiek} access {setip}[_k_]/{in_clrip}[_k_] and {setie}[_k_]/{clrie}[_k_], respectively.
```
christian-herber-nxp, line 292: As this mapping was removed, I believe we are missing the functional specification for how the pending bits get asserted etc.

christian-herber-nxp (changes requested):
I am in principle fine with these changes.
I do think eidelivery should stay as it was.
While technically this allows dynamic selection between ACLIC and IMSIC, 
1. I do not see a technical problem with that
2. The register is WARL, so ACLIC systems will likely have ACLIC as the only legal option.

My other concern is that there might now be gaps in the specification, as you pulled in register definitions from APLIC, but it is not clear to me that the behavior of APLIC is also required.

src/aclic.adoc
```diff
@@ -252,72 +242,62 @@ Of the main ACLIC goals, these targets are addressed here:
...
-If Smaclic or Ssaclic extensions are present, reset initializes {eidelivery} to 0x20000000.
+Since the ACLIC is closely coupled to a single hart for which AIA already defines an
+indirect CSR access mechanism, interrupt source and priority configuration is performed 
+using new indirect CSRs instead of the memory-mapped registers of an external APLIC. 
+The source and priority configuration data follow the same definitions as the equivalent APLIC fields. 
 
-When interrupts are delivered from an ACLIC, the following behavior is valid:
+Note: 
+  * A hart cannot have more than one closely coupled external interrupt controller; 
+  * in particular a hart cannot have both an ACLIC and an IMSIC.
```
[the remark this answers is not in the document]
mark-honman: So - "a hart can have exactly one closely coupled external interrupt controller" ?

src/aclic.adoc
mark-honman: eidelivery is a register in the interrupt file of a hart-connected interrupt controller. It determines whether the result of the interrupt file's filter tree goes to the external interrupt input of the hart's major interrupt controller corresponding to the mode of the interrupt file. AIA provides a "special case" where the external interrupt input is connected directly to the output of an APLIC domain. That's all still supported. 
My understanding is that we don't need an extra eidelivery bit because there is no possibility for IMSIC and ACLIC to coexist, because they occupy the same CSRind space.

src/aclic.adoc
mark-honman: Noted... I think we should specify that directly or perhaps by analogy with APLIC.

src/aclic.adoc
```diff
@@ -326,15 +306,21 @@ When XLEN = 32, each `xireg` register controls the ACLIC priority setting of fou
 [%autowidth]
 |===
 | `xiselect` |  `xireg` size |  `xireg` state
-| 0x1000     |   4B          | `xacliciprio0`
+| 0x100      |   4B          | `eiprio0`
```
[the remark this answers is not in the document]
mark-honman: The CLIC-specific CSRind base of 0x1000 goes back to the time that CLIC was an independent extension.
With ACLIC being in the AIA family, if the CSRind space from 0x100 - 0x1FF is available it would make more sense for all indirect CSRs that relate to an interrupt file to occupy a contiguous block of CSRind space.

src/aclic.adoc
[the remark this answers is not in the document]
christian-herber-nxp: I meant to say APLICs can be cascaded.

For CLINT, just avoid this. Rather, state that the external controller is connected as an input to XEIP.

christian-herber-nxp: But then you would still have to change the meaning of the IMSIC e delivery value to be IMSIC or ACLIC

src/aclic.adoc
```diff
@@ -252,72 +242,62 @@ Of the main ACLIC goals, these targets are addressed here:
...
-For XLEN=64, {eipk} and {eiek} only exist for even numbered indices of _k_. In this case,
-{eipk} and {eiek} access {setip}[_k_]/{in_clrip}[_k_] and {setie}[_k_]/{clrie}[_k_] in the lower half of the CSR,
-and {setip}[_k+1_]/{in_clrip}[_k_] and {setie}[_k+1_]/{clrie}[_k_] in the upper half of the CSR, respectively.
-
-NOTE: Mirroring these APLIC registers at a core level has several advantages.
-  * It removes unnecessary resource duplication between IMSIC and APLIC when integrated with a single hart.
+NOTE: Implementing these APLIC-equivalent registers as core level indirect CSRs has several advantages.
   * It simplifies the address decoding logic within the interrupt controller by using CSR access over memory mapped ones.
-  * It simplifies the handling of level-sensitive interrupts, avoiding additional checking of the pending flag in the APLIC in a handler of a level sensitive interrupt.
+  * It simplifies the handling of level-sensitive interrupts, avoiding the need for a separate pending flag as in the APLIC.
```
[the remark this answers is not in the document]
christian-herber-nxp: I think it is good that we document this as a reason why this solution is preferred over using IMSIC+APLIC.
We just have to find the right place

src/aclic.adoc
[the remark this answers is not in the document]
mark-honman: OK

src/aclic.adoc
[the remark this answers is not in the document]
mark-honman: How about replacing it with a bullet
* It simplifies software, because CSR operations are atomic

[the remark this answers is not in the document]
christian-herber-nxp: that is a separate point.
What the bullet in question means is the following:
With APLIC + IMSIC, The APLIC automatically clears interrupts when sending to the IMSIC
For level based interrupts, a handler has to check the APLIC to see if the level interrupt is actually no  longer pending.

src/aclic.adoc
mark-honman: AIA section 4.7 Precise effects on interrupt-pending bits describes pending bit behaviour for all of the APLIC source configurations. This can be used as the basis of an interrupt-pending behaviour description for ALIC.

src/aclic.adoc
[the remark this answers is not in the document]
mark-honman: The bullet can be changed to
"It does not require the special considerations for level-sensitive interrupts that are needed when APLIC delivers interrupts to an IMSIC"

That is a reference back to an AIA section "special considerations for level-sensitive interrupts", anyone interested can go read the details there...

src/aclic.adoc
mark-honman: IMO we should keep the original AIA eidelivery definition until we have a specification of how ACLIC and IMSIC could possibly be attached to the same hart. In the meantime I will draft a Smaclicmsi extension...

src/aclic.adoc
```diff
@@ -253,30 +262,29 @@ indirect CSR access mechanism, interrupt source and priority configuration is pe
...
+{eipk} behaviour depends on the interrupt source configuration, and is detailed below. 
 
-The number of interrupt identities supported by an ACLIC interrupt file is one less that a multiple of 64,
-unless the number of identities is less than 64. 
-Unlike the IMSIC for which the minimum number of supported identities is 63,
-an ACLIC can additionally support 7, 15, 31, or 63 interrupt identities. 
+The number of interrupt identities supported by an ACLIC interrupt file is one less than a multiple of 64.
+An ACLIC implementation can implement fewer than 63 interrupt identities.
 In this case the {eip} and {eie} bits and source and priority configuration for the unimplemented
 identities are WARL 0.
```
christian-herber-nxp, line 280: is there a difference between WARL 0 and roz?

christian-herber-nxp (commented):
What is written looks fine to me. 
I would like to give it a deeper review front to back to make sure nothing is missing.

PR #630, opened 2025-09-27
Smtp: Major and minor identity in a0+a1 for external irqs

### LINKED ISSUE #628: Tpush should pass major and minor interrupt identity to handlers
Currently, an external interrupt receives the minor identity in a0. This works for HW-vectored interrupts. 

For SW vectored interrupts, it is important to also pass the major identity, so that the dispatcher can call the driver of the external controller.

I propose to change external interrupts to receive in
a0: major identity
a1: minor identity

christian-herber-nxp:


christian-herber-nxp:
I am actually considering placing xcause in a0 instead to have consistency between interrupts and synchronous exceptions.
Otherwise, you would need a different behavior between vectored and non-vectored interrupts, which is an additional complication.

PR #661, opened 2025-12-02
Smaclic: Improve introduction to ACLIC

### LINKED ISSUE #615: Clarify introduction to ACLIC
Currently, section 3 says "The ACLIC is a tailored APLIC". 
That steered my mental model in the wrong direction - not helped by the detailed description of ACLIC behaviour in section 3.1. 

A possible better description, starting with the line "To acheive this..." :

ACLIC is a hart-local external interrupt controller, optimized for resource efficiency. ACLIC uses the AIA defined CSR and indirect-CSR interfaces for hart-local interrupt controllers, which are also  implemented by the AIA IMSIC. ACLIC adds indirect CSRs that enable wired interrupts to be configured using APLIC concepts that are relevant to directly integrated interrupt controllers. 

In terms of compatibility, the goal is that a hart with Smaclic/Ssaclic extensions can be used at with the same trap handlers as one developed for AIA.

The configuration of interrupt delegation, source mode, and interrupt priority work the same way as in APLIC, but use the indirect-CSR interface introduced by AIA for hart-local controllers.

christian-herber-nxp:


christian-herber-nxp:
@mark-honman : As I have worked through your other issue, I feel that the main issue here was not steering your mental model to APLIC, but rather not steering your mental model hard enough to APLIC.

mark-honman (approved):
Very helpful

PR #662, opened 2025-12-02
Rework aclic structure

christian-herber-nxp:


src/aclic.adoc
```diff
@@ -196,116 +200,117 @@ The table below provides a summary of the ACLIC extensions.
...
-A minimal implementation of the AIA `mvip` CSR is suggested by the AIA specification: +
-"When supervisor mode is implemented, the minimal required implementation of mvien and mvip has all bits being
-read-only zeros except for mvip bits 1 and 9, and sometimes bit 5, each of which is an alias of an existing 
-writable bit in mip. When supervisor mode is not implemented, registers mvien and mvip do not exist."
+An ACLIC is made up of three components:
 
-=== New Interrupt Delivery Mode
+. A hart-local APLIC (Non-ISA)
+. Indirect CSR access to APLIC registers (Smaplicind, Ssaplicind)
+. Interrupt control interface based on IMSIC (Smictrl, Ssictrl)
```
mark-honman: How about:
. hart-local APLIC domains (Non-ISA)
. IMSIC-based indirect CSR physical control interface to hart-local APLIC domain state

Or make the second point shorter "indirect CSR physical control interface to hart-local APLIC domain state" and later explain that the Indirect CSR control interface extends what AIA defines for IMSIC.

src/aclic.adoc
```diff
@@ -196,116 +200,117 @@ The table below provides a summary of the ACLIC extensions.
...
 
-The Incoming Message Signaled Interrupt Controller (IMSIC) contains the {eidelivery} CSR,
-which defines the interrupt delivery mode.
-A new delivery mode is defined:
+ACLIC at machine level is enabled by Smaclic,
+which is a group of extensions composed of Smaplicind and Smictrl,
+and implies the implementation of a hart-local APLIC.
```
mark-honman: "implementation of hart-local APLIC domains" ?

src/aclic.adoc
```diff
@@ -196,116 +200,117 @@ The table below provides a summary of the ACLIC extensions.
...
+  CSR-based interfaces always use the same CSR register numbers
+
+No dedicated access to interrupt pending or enable bits is provided in this extension.
+Access to these bits is expected to be provided in other extensions.
 
 To have full control over the necessary APLIC registers, the following indirect CSR access is added.
+The Smaplicind extension gives machine-level access to the root domain (x=m),
+and Ssaplicind gives supervisor-level access to the child domain (x=s)
 
 ==== Interrupt Priority (iprio)
```
mark-honman: Previously in the ACLIC spec, there was a reference to APLIC target registers. That's a good idea, so we should probably just call the indirect CSR "target", and it has an iprio field. 

Now the controversial idea about sourcecfg and target, which I don't think should be reflected in this PR. 
That is, provide one xiselect slot per interrupt id, and have sourcecfg[i] in xireg and target[i] in xireg2. 
Downsides are a bit more area from decoding the address, and consumption of more CSRind address space. 
Advantages are that it's much more like APLIC sourcecfg and target, and it's potentially easier for SW to initialize the control information for a specific interrupt ID.

src/aclic.adoc
```diff
@@ -196,116 +200,117 @@ The table below provides a summary of the ACLIC extensions.
...
 
-=== Indirect CSR access to APLIC registers
+The `IE` bit is fixed to 1, as its functionality is subsumed by the {xstatus}.`xIE` bits at the hart.
+The `BE` bit is now reserved.
+Endianess is controlled exclusively by the setting in the {xstatus} register at the hart.
+
+Further simplifications are:
+
+* The Child Index field in the sourcecfg[i] register is always '1' when D=1 (i.e. delegation to child domain). This is only possible in Machine mode. In Supervisor mode, D=0, as there is no child domain.
+* As the ACLIC is not MSI capable, xmsiaddrcfg[h] and genmsi registers are not required.
```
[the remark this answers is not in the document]
christian-herber-nxp: coexistence with IMSIC is not necesscerily the key point here. ACLIC is a wired interrupt controller. So no MSI generation is required, whoever the consumer of the MSI might be.

mark-honman (commented):
Not much to say here (thankfully)... a couple of typos some of which might be mine. 

One thing I am still wondering about - pros/cons of using the term "APLIC domain" vs "AIA interrupt domain" (because "interrupt domains" is the actual term used in AIA). 

Definitely stick with APLIC domains" in this PR because it tells implementors that the behaviour of an ACLIC domain is almost identical to an APLIC domain.

src/aclic.adoc
mark-honman: Strange to see it go, but makes sense.

PR #696, opened 2025-12-30
Smnip: Simplified definition of threshold, fully relying on IPRIO

christian-herber-nxp:


src/aclic.adoc
```diff
@@ -734,22 +734,23 @@ The reset behavior of other fields is platform-specific.
...
 when the following condition holds:
 
 [source]
 ----
 IPRIO >= xithreshold & NIPPRIO_MASK
+----
 
-where IPRIO is equivalent to the value reported in xtopi CSR, i.e.
+where `IPRIO` is equivalent to the value reported in {xtopi} CSR,
+if the interrupt source is the highest priority pending-enabled one.
```
mark-honman: A slight reword would make this clearer

"where IPRIO is the value that would be reported in the {xtopi} CSR if the interrupt source is the highest priority pending enable one"

src/aclic.adoc
```diff
@@ -734,22 +734,23 @@ The reset behavior of other fields is platform-specific.
...
-1 < iprio[i] <= 255: IPRIO = iprio[i]
-255 < iprio[i]: IPRIO = 255
-iprio[i] = 0:
-  default priority higher than a machine external interrupt: IPRIO = 0
-  default priority lower than a machine external interrupt: IPRIO = 255
-----
+[NOTE]
+====
+IPRIO is derived from the configured priority or default priority values
+according to the rules defined in AIA.
```
evgeniy-paltsev: There is no `default priority values`. AIA defines `default priority order`; however it is not a numerical value which can be used for IPRIO.

src/aclic.adoc
```diff
@@ -734,22 +734,23 @@ The reset behavior of other fields is platform-specific.
...
 
-1 < iprio[i] <= 255: IPRIO = iprio[i]
-255 < iprio[i]: IPRIO = 255
-iprio[i] = 0:
-  default priority higher than a machine external interrupt: IPRIO = 0
-  default priority lower than a machine external interrupt: IPRIO = 255
-----
+[NOTE]
+====
+IPRIO is derived from the configured priority or default priority values
```
christian-herber-nxp:
```suggestion
IPRIO is derived from the configured priority or default priority order
```

PR #702, opened 2026-01-06
698 smtp add sstp extension which add an s mode equivalent return instruction

christian-herber-nxp:


src/aclic.adoc
```diff
@@ -418,6 +418,7 @@ and the conditions for a virtual instruction exception apply,
 in which case a virtual instruction exception is raised
 when in VS or VU mode instead of an illegal instruction exception.
 
+[[smcsps, Smcsps]]
```
mark-honman: Not sure if this is relevant because I don't know what [[ ]] is supposed to do; but should the bracketed list also include Sscsps, sscsps ?

src/aclic.adoc
```diff
@@ -492,20 +494,20 @@ An example sequence with one vertical and one horizontal trap will look like thi
...
+M     mcspspop         0         0          n
 M     mret             0         0          n
 M     restore context  0         1          n
-M     cspspop          1         0          y
+M     mcspspop         1         0          y
 S     mret             1         0          n
 ----
 
 The stack pointer is swapped when changing from S-mode to M-mode,
 and is swapped back when returning to S-mode.
```
mark-honman: Maybe re-word this because when there is a mode change, it's not actually a stack pointer swap. Suggestion:

When changing from S-mode to M-mode, the stack pointer is stored into ssp and loaded from msp, without affecting ssp.PUSH and ssp.PPUSH. 
The reverse happens when changing from M-mode to S-mode. 

I'm not convinced that what I wrote about preserving PUSH and PPSH is correct...

PR #703, opened 2026-01-06
Smip: Adding note stating incompatiblity with shared entry points for…

christian-herber-nxp:
… exceptions and interrupts

### DIFF
--- a/src/aclic.adoc
+++ b/src/aclic.adoc
@@ -1066,6 +1066,11 @@ It performs the following operations in order:
 NOTE: Intentionally, the GPRs are saved on the stack owned by the interrupted context.
 This is done to simplify and accelerate OS thread switching.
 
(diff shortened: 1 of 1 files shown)


src/aclic.adoc
```diff
@@ -1066,6 +1066,11 @@ It performs the following operations in order:
 NOTE: Intentionally, the GPRs are saved on the stack owned by the interrupted context.
 This is done to simplify and accelerate OS thread switching.
 
+NOTE: As this extension modifies interrupt trap behavior but not keep synchronous exceptions traps,
```
mark-honman: Sorry I'm late to review, there are some typos here. 

Suggested wording:

NOTE: As this extension modifies interrupt trap behavior but not the behaviour of synchronous exception traps,
it should only be used in interrupt handling modes that provide separate entry points for 
interrupt and synchronous exception traps.
Specifically, mtvec.mode = 0 (direct) is incompatible.

PR #727, opened 2026-01-28
Smcsps: Making instruction illegal when MSBs are not zero

### LINKED ISSUE #719: *cspspop *cspspush constraints/assumptions documentation and generalization/enforcement
This addresses both the Machine and supervisor  versions of mcspspush and mcspspop.

First is that sp 16 byte alignment is an absolute requirement throughout user mode code and not just  a calling convention.

Supervisor mode code can mitigate this restriction on its own code through disabling interrupts.

In the case user code has for whatever reason, set sp to a non-16 byte aligned value and is interrupted, on return from the handler the value of sp is changed, the low 4 bits are cleared.

This could cause errant behaviour in the user code.
At minimum a non-conformance to the requirements of the push/pop instruction is quietly ignored.

Minimally, this requirement and behaviour should be clearly documented.

There are ways of mitigating these concerns that I would like to propose so that this feature may be more generalized and thus applicable to more applications.

1) trap on illegal value (or other appropriate code) if sp has non-zero low 4 bits. (especially for scspspush.
  The current non-trapping behaviour is obtained by explicitly clearing the low 4 bits before the *scspspush instruction. (  andi sp, sp, -16 )

2) on interrupt the low 4 bits of sp are saved into *pistatus .
    Software can recover the user sp low 4 bits before a return (along a slow path, after handling the interrupt).

   Both of these allow for the minimal detection of mandated sp alignment. 

3) establish a mode in U/S that loads/writes to sp with non-zero low 4 bits trap.
    This has the benefit of enforcing the alignment mandate and identifying where the code becomes non-compliant.

christian-herber-nxp:


christian-herber-nxp:
@David-Horner please review

PR #730, opened 2026-01-28
Smip: specified fault handling behavior

christian-herber-nxp:


src/aclic.adoc
```diff
@@ -1182,6 +1182,10 @@ It performs the following operations in order:
   * Adjusts the stack pointer to create the stack frame
   * Conditionally performs a stack pointer swap by executing cspspush sp
 
+When creating the stack frame,
+the stack pointer is adjusted the the next 128b-aligned valued
+that is sufficient to save the specified registers.
```
evgeniy-paltsev: If we just adjust SP to the next 128b-aligned value, then it means that we may change SP during the interrupt handling (the SP before the interrupt and the SP after the interrupt may be different).

christian-herber-nxp: as the stack frame we create does not participate in any ABI compliant calls, it might actually not be required to align it.

Issue #726, opened 2026-01-28
Exception traps may break interrupt nesting logic with Smnip & Ssnip extensions


evgeniy-paltsev:
If we do not add Smip & Ssip extensions - we use standard mret/sret instructions in both interrupt and exception handlers.

However, Smnip & Ssnip extensions add new behavior to mret/sret instructions - they update xpistatus.pithreshold from xithreshold on mret/sret execution.

So, if we get exception trap during interrupt handler execution - we'll return with modified xithreshold and break interrupt nesting logic.

reosdev:
Proposed [fix](https://github.com/riscv/riscv-fast-interrupt/pull/732) introduces consequent issues:
- new ACLIC specific status CSR must be threated as part of all traps context;
- as the exceptions cannot rely on subset of status bits in ACLIC custom status CSR, the stack frame must duplicate status (custom ACLIC and regular status CSRs);
- triggering CSR-to-CSR copy (xpistatus->xithreshold)  by return instruction is not aligned with RISC core pipeline and either requires special support in core logic, or adds few cycles to each return.
- write to threshold in most cases will lead a pipeline stalls (otherwise core may trap on request asserted with wrong threshold;
- compatibility with the software is broken, as from now developer must operate with multiple status registers before executing xret, which is not always performed from handler; this is not atomic procedure and must be wrapped into critical section with interrupts disabled globally.
- disabling interrupts globally does not protect from traps to higher privilege level, which may confuse software in determining proper trap context.

Above are only few findings that demonstrate risks of quick fixes provides without context consideration (just few words changed in PR).

christian-herber-nxp:
I will follow up on Monday. Have a good weekend

PR #732, opened 2026-01-29
Smnip: fixing situation in which an exception leads to the interrupt …

christian-herber-nxp:
…threshold being wrong


reosdev:
Quick analysis shows that requiring interrupt threshold manipulation in exception trap context may be dangerous. Please see the explanation in https://github.com/riscv/riscv-fast-interrupt/issues/726#issuecomment-3818186588

PR #757, opened 2026-02-03
Fix: Remove misleading statement regarding H-extension scope

### LINKED ISSUE #731: Misleading statement regarding H extension
Currently there is a phrase in the spec (Version v0.17):

> Future specifications could extend the scope of these extensions to also be applicable to systems with the H extension.

From my point of view - the phrase is misleading and it should be rephrased or removed entirely.
Current design does not allow efficient virtualization. If it is used in H-extension systems - we won't get performance/latency benefits. It's quite opposite - we'll get degradation in both VM switch and interrupt handling.

DymShanks:
Removes the statement: "Future specifications could extend the scope of these extensions to also be applicable to systems with the H extension."

As discussed in Issue #731, this statement implies a future path for H-extension support that is architecturally impractical for single-hart designs. Removing it prevents misleading implementers.

Ref: Discussion with @reosdev and @evgeniy-paltsev.
Fixes #731

### DIFF
--- a/src/aclic.adoc
+++ b/src/aclic.adoc
@@ -197,7 +197,6 @@ The table below provides a summary of the ACLIC extensions.
 |===
 
 As this specification aims for single-hart systems, no support for the H extension is foreseen.
-Future specifications could extend the scope of these extensions to also be applicable to systems with the H extension.
 
 NOTE: The extensions defined here are orthogonal to the NMI and RNMI
 mechanisms. Their behavior is unchanged by the extensions of ACLIC.


reosdev:
I don't think it is good to close the bugs without actual PR review. I see 5 references to H-extension:

1. Chapter 1 (line 114-116):
> "no support for the H extension is foreseen. Future specifications could extend the scope..."
2. Section 2.2.2 State Enable (line 282-284):
> "except if the hypervisor extension is implemented... virtual instruction exception is raised when in VS or VU mode"
3. Section 3.3 State Enable (line 348-350):
> Same text about hypervisor extension and VS/VU mode exceptions
4. Section 4.5 State Enable (line 582-584):
> Same text about hypervisor extension and VS/VU mode exceptions
5. Section 5.1.5 State Enable (line 700-702):
> Same text about hypervisor extension and VS/VU mode exceptions

Proposed PR fixes just first one.

christian-herber-nxp:
The issue this PR was related to was clearly about the line in chapter one, not the other ones.

PR #794, opened 2026-03-06
Use ghcr.io/riscv/riscv-docs-base-container-image:latest

rpsene:
This updates the docs container image reference in the root Makefile.

The old `riscvintl` image has been replaced with the `ghcr.io/riscv` image.


wmat (approved):
LGTM

PR #823, opened 2026-04-07
Smcsps: Clarify that stack pointer swapping always operates on x2

christian-herber-nxp:
@pkennedyTT please review


christian-herber-nxp:
replaced by https://github.com/riscv/riscv-fast-interrupt/pull/828

PR #825, opened 2026-04-14
816 issue smnip guard changed trap behavior of smnip with state bits and add handler examples

### LINKED ISSUE #816: [Issue]: Smnip: Guard changed trap behavior of Smnip with state bits
### Document Version

0.18

### Current Specification

Currently, Smnip changes the trap behaviorin two ways:

When a trap is taken into level x, the current value of xithreshold is written to xpistatus.pithreshold.

Additionally, if the trap was taken on an interrupt, xithreshold is set to IPRIO. This automatically raises the interrupt-enable threshold.

### Problem Definition

This is a behavior change that is not backwards compatible with the existing specification.

### Proposed Solution

The behavior should be enabled from a state bit.
The first part, 

> When a trap is taken into level x, the current value of xithreshold is written to xpistatus.pithreshold.

Should be guarded by a bit introduced in Smnip. 

For the second part:

> Additionally, if the trap was taken on an interrupt, xithreshold is set to IPRIO. This automatically raises the interrupt-enable 
> threshold.

As the automatic behavior only benefits in interrupt vector table mode, couple the behavior to xtvec.mode=11 (IVT mode).

christian-herber-nxp:


christian-herber-nxp:
replaced by https://github.com/riscv/riscv-fast-interrupt/pull/828

PR #832, opened 2026-05-19
Smip: Removing hard dependency on Smcsps

### LINKED ISSUE #791: why is it specified that the Smip extension depends on the Smcsps extension?
The manual states that the Smip extension depends on the Smcsps extension.
I thought about it and it seems that automatic stacking could also be implemented with just one stack, meaning that the Smip extension could be implemented even without supporting the Smcsps extension. So why is it specified that the Smip extension depends on the Smcsps extension?

christian-herber-nxp:
@XIVN1987: Sorry for the long turn around time on this issue.
I did remove the dependency on Smcsps, as it is not necessary.

### DIFF
--- a/src/aclic.adoc
+++ b/src/aclic.adoc
@@ -1838,8 +1838,6 @@ this exception cannot occur.
 [[smip, Smip/Ssip]]
 === Interrupt handler push and pop extension (Smip & Ssip)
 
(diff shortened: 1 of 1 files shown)


christian-herber-nxp:
@XIVN1987 I will merge this. If you get to this at a later point in time, let me know in case this did not address your issue.

Issue #742, opened 2026-02-02
False claim about simplified level-sensitive interrupt handling


reosdev:
> Providing access to APLIC pending and enable bits at the hart-level has several advantages over using an IMSIC with interrupt delivery from an APLIC
> It simplifies the handling of level-sensitive interrupts, avoiding additional checking of the pending flag in the APLIC in a handler of a level sensitive interrupt.

The claim that eipK/eieK iCSR interface "simplifies" handling is incorrect. Compare clearing a pending bit by interrupt source number:

**ACLIC eipK iCSR approach (RV32):**
```asm
# a0 = interrupt source number
    srli  t0, a0, 5        # t0 = K = N / 32 (register index)
    andi  t1, a0, 0x1f     # t1 = bit position = N % 32
    li    t2, 0x80         # eip0 base in xiselect space
    add   t0, t0, t2       # t0 = xiselect value for eipK
    csrw  miselect, t0     # select eipK register
    li    t2, 1
    sll   t2, t2, t1       # t2 = bit mask
    csrc  mireg, t2        # clear the bit
# 8 instructions
```

**APLIC clripnum MMIO approach:**
```asm
# a0 = interrupt source number
# a1 = APLIC domain base address
    sw    a0, CLRIPNUM_OFFSET(a1)  # write source number
# 1 instruction (+ base address setup if not cached)
```

The MMIO clripnum interface is dramatically simpler: one store vs eight instructions with bit manipulation. The eipK interface adds complexity, not simplicity.

PR #836, opened 2026-05-26
Smijt: Added definition how to abort an interrupt

### LINKED ISSUE #721: Fault handling on HW stacking is broken
As per ACLIC spec:

> If synchronous exceptions can occur during the context save/restore segments, handling can only be resumed if the double trap extension is supported.

However, it's not clear how double trap can help here. Even if we handle the second exception trap with double trap - where should we return?
We were in the middle of the HW stacking sequence.

Current spec (Version v0.17) completely missing the details how it should work.

### LINKED ISSUE #724: Missing order between the trap entry operations from different extensions
With multiple extensions enabled we have multiple operations during trap entry which can't be done atomically in one step:

Smivt, Ssivt:
Obtainin interrupt IIDs for vectoring, vector jump

Smip & Ssip:
Register saving, SP adjustment, SP swap.

Smnip & Ssnip:
Updating xpistatus.pithreshold, xithreshold.

Currently there is no order defined between the operations from different extensions.

### LINKED ISSUE #725: Spec missing info about M-mode interrupts behavior during S-mode HW stacking sequence
The current spec (Version v0.17) missing information about what happens if we have ongoing S-mode HW stacking (because of S-mode interrupt trap) and M-mode interrupt becomes pending.

Do we block M-mode interrupts until S-mode HW stacking is completed?

christian-herber-nxp:
Fixes #721, Fixes #724, Fixes #725

### DIFF
--- a/src/aclic.adoc
+++ b/src/aclic.adoc
@@ -1993,6 +1993,14 @@ Explicitly, {xepc}, {xistatus} remain unchanged.
 NOTE: In the typical case of tail chaining, the check for another pending enabled interrupt is at the start of the {xipopret} instruction.
 In this case, the entire context save and restore sequence can be skipped.
 
+==== Aborting Interrupt Traps
+
+The requirements on how to abort an interrupt in <<interrupt-trap-abort>> also apply to this extension.
+
+In idempotent memory regions,
+the store operations of an aborted interrupt trap may have updated memory.
+Non-idempotent memory regions must not have been updated during the aborted interrupt trap.
+
 ==== Interrupt Fault Handling
 
 This extension introduces additional operations during an interrupt trap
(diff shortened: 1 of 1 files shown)


evgeniy-paltsev:
Hi @christian-herber-nxp,

#742 is unrelated to the changes in this PR.

Either the PR has changed and is no longer related, or there may be a typo in the issue number.

christian-herber-nxp:
> Hi @christian-herber-nxp,
> 
> #742 is unrelated to the changes in this PR.
> 
> Either the PR has changed and is no longer related, or there may be a typo in the issue number.

you are right. dont know what happened. I have updated to the issues which i believe are related

evgeniy-paltsev (commented):
Hi, I've checked the changes and following is not clear for me:

We add explanation for the jump table extension. However, the linked issues are also valid for the HW stacking extension. For it we do alter the architecture state before we start to save values on a stack. How is it expected to operate fo this case?

christian-herber-nxp:
> Hi, I've checked the changes and following is not clear for me:
> 
> We add explanation for the jump table extension. However, the linked issues are also valid for the HW stacking extension. For it we do alter the architecture state before we start to save values on a stack. How is it expected to operate fo this case?

It is supposed to operate the same way, and the hardware-stacking extension refers to the Smijt extension on this topic.

christian-herber-nxp:
I have updated the Smnip section to include also the section on aborting interrupts to be as explicit as possible.

PR #853, opened 2026-08-13
Smip/Ssip: Fix stack frame size description

### LINKED ISSUE #852: [Issue]: Smip/Ssip RV64 stack frame size is not minimal as stated
### Document Version

ACLIC spec v0.20

### Current Specification

The current Interrupt handler push and pop extension (Smip & Ssip) [states that:](https://github.com/riscv/riscv-fast-interrupt/blob/master/src/aclic.adoc?plain=1#L1699-L1701)

> The size of the stack frame created is set to XLEN bytes,
which is the minimum stack frame size that can hold the six pushed registers,
and is psABI-compliant.

### Problem Definition

This is true for RV32I:
- Six registers require `6 * 4 = 24` bytes.
- The [standard psABI requires the stack pointer to remain 16-byte aligned](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/76b837ec964509f4bac11c66e5d7106b6a1e626a/riscv-cc.adoc?plain=1#L268-L281).
- The minimum aligned frame is therefore 32 bytes, equal to `XLEN bytes`.

However, it does not appear to be true for RV64I:
- Six registers require `6 * 8 = 48` bytes.
- 48 bytes is already a multiple of the 16-byte psABI stack alignment.
- So the minimum aligned frame is 48 bytes, while `XLEN bytes` is 64 bytes.

The 64-byte frame is sufficient and psABI-compliant, but it is not the minimum frame size.

### Proposed Solution

The history in PR #828 suggests that fixing the frame size to `XLEN bytes` was intentional.

So maybe we could just replace the "minimum" wording:

> The size of the stack frame created is set to XLEN bytes, which is sufficient to hold the six pushed registers and is psABI-compliant.

zeyi2:
This patch changes Smip/Ssip stack frame size from XLEN bytes to:

- 24 bytes on RV32E
- 32 bytes on RV32I
- 48 bytes on RV64

Fixes #852

### DIFF
--- a/src/aclic.adoc
+++ b/src/aclic.adoc
@@ -1696,9 +1696,9 @@ It performs the following operations in order:
 NOTE: The stack pointer swap is performed as the first action,
 as the integrity of the sp of a less privileged level cannot be guaranteed at all times.
 
-The size of the stack frame created is set to XLEN bytes,
-which is the minimum stack frame size that can hold the six pushed registers,
-and is psABI-compliant.
+The size of the stack frame created is set to 24 bytes on RV32E, 32 bytes on RV32I
+and 48 bytes on RV64, which are the minimum stack frame sizes that can hold the
+six pushed registers, and are psABI-compliant.
 
 NOTE: It is the responsibility of the programmer to ensure
 that sp contains a psABI-compliant stack pointer value


zeyi2:
I'm not sure whether specification wording issues like this should be addressed directly through a PR. Please let me know if a different process would be more appropriate.

christian-herber-nxp:
Thanks for the PR.
The stack size for RV64 only needs to be 48 B. Requiring XLEN=64 bytes is overkill. 
We could formulate this as a formula, but it might just make most sense to call out the frame sizes explicitly.

zeyi2:
> Thanks for the PR. The stack size for RV64 only needs to be 48 B. Requiring XLEN=64 bytes is overkill. We could formulate this as a formula, but it might just make most sense to call out the frame sizes explicitly.

Thanks for reviewing!

One follow-up question: I just noticed that the repository [CI builds examples for RV32E/ILP32E](https://github.com/riscv/riscv-fast-interrupt/blob/6470634f3ab79aa21def816235fce92e6cc13382/.github/workflows/handlers.yml?plain=1#L38-L41). Since [ILP32E only requires 4-byte stack alignment](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/01017c343cd6d89ed4d1d568b1c75fac79d2a689/riscv-cc.adoc?plain=1#L668-L673), six 32-bit registers require 24 bytes, so 24 bytes would be the minimum frame
size in that case.

Should the specification distinguish between RV32E (24 bytes) and RV32I (32 bytes)?

For now, I plan to update the PR using these the explicit frame sizes. If that is not the intended behavior, I'll be happy to adjust it accordingly.

christian-herber-nxp:
> Should the specification distinguish between RV32E (24 bytes) and RV32I (32 bytes)?

We did not consider this an important enough use case to spend additional configuration space on. 
If the stack size overhead is paramount, you can deactivate Smip.

PR #860, opened 2026-09-01
Smijt: fixing access to fields in xijt (WARL/WPRI)

christian-herber-nxp:
@jrahmeh please review


jrahmeh:
The EVH field type looks good.  The width is still 2. Should that be 1?

christian-herber-nxp:
> The EVH field type looks good. The width is still 2. Should that be 1?

the physical width of the box in the rendering is 2. That is needed to fit the text, but the denoted width should be 1.

PR #867, opened 2026-09-11
Tagged normative rules

christian-herber-nxp:


rpsene:
Hi @christian-herber-nxp, a status update on this PR.

**The `build` failure was not caused by your changes.** The weekly rebuild of `riscv-docs-base-container-image` on 2026-09-14 pulled in json 3.0, which broke asciidoctor-diagram in the HTML build (`unknown keyword: create_additions`). That is fixed in riscv/riscv-docs-base-container-image#24, and a new `latest` image is published. I re-ran the `build` check here and it now passes, including the normative rule generation from your second commit (151 rules written to JSON and HTML).

**The only remaining blocker is DCO.** GitHub still shows the head of this PR as `b45847a`, pushed at 14:52 UTC today, and that commit has no `Signed-off-by:` trailer (the first commit, `7c76723`, has one). If you already added the sign-off, it looks like it has not reached this branch yet. From the `normative-rule-tagging` branch:

```
git commit --amend -s -S --no-edit
git push --force-with-lease
```

`-S` keeps the commit signed, which the repository also requires. Once that is pushed, the DCO check should turn green.

